PR #397, opened 2015-02-10
Build conversion to new netflixoss plugin

rspieldenner:


rspieldenner:
The 2.0 branch should be renamed 2.x or 2.0.x after this is merged

PR #417, opened 2015-08-19
waitToComplete properly

wy96f:
AbstractRestore will wait forever if exception is thrown during downloading file. Fix it by using future returned by executor.submit() function.


tulumvinh:
Hello,

We will review the pull request.

Thank you

PR #419, opened 2015-08-19
finally shutdown executor in AbstractRestore

wy96f:
Finally shutdown executor in AbstractRestore so that Restorer can exit.


tulumvinh:
Hello,

We will review the pull request.

Thank you

PR #422, opened 2015-09-17
2.x

scottmcmaster:


tulumvinh:
Hi Scott,
Are you finding that type parameter is required for compilation?  I asked as I found that adding the type parameter in this class causes Guice to not bind implementation to interface correctly.

Regards

scottmcmaster:
Hi. Yes, compile errors at lines 77 and 126. The calls to filteredRemote use an unchecked conversion, which is ok in-and-of itself. But then that causes the return value to be type-erased to List<Object>, which the foreach's don't approve of.

Hmm, I only added the type parameter to the private field, not to the @Inject'd parameter, so I'm not sure why Guice would have any opinion in this case. Can you let me know the scenario to reproduce your binding issue (since the tests pass anyhow)? Alternatively, it should also be possible to work around this by changing those two foreach's to regular for's.

Thanks!

tulumvinh:
Hi Scott.

Follow-up questions/clarifications.

1.  I am not getting the compile errors.  In addition, the Cloudbees' jobs (https://netflixoss.ci.cloudbees.com/job/NetflixOSS/job/Priam/job/Priam-2.x-release/5/console and https://netflixoss.ci.cloudbees.com/job/NetflixOSS/job/Priam/job/Priam-2.x-snapshot/4/console) passed.

These build use idk 1.7.  Is your idk > 1.7?

2. Yes, you're right.  adding the type parameter to the constructor did confuse Guice.  Adding the type parameter to the private member should be fine.

Lets close the loop on my question #1 and then I'll merge your changes,

Regards

scottmcmaster:
Sorry, yes, I'm using an OpenJDK Java 8, and to really close the loop, I believe the issue in question is discussed here: http://mail.openjdk.java.net/pipermail/compiler-dev/2013-October/007726.html

Thanks again!

Issue #431, opened 2015-10-27
Priam is not deleting incremental files.


sciclon:
Hi guys,

After investigating I realized that Priam is uploading the incremental backups to the S3 buckets and the snapshot as well but I see the files still remains in /backup/ folder.

I know they are hard links but they are not deleted from the backup folder, they are referenced just once (it is not because the file exists in data folder).

My doubt is, should I delete the files on my own? if this is true, should I compared what was uploaded to S3 and then remove the files? 

Additional info:
Cassandra version: 2.0.9 ( I know is a bit older )

Priam log:
08/SST/system/compactions_in_progress/system-compactions_in_progress-jb-4813-Filter.db with chunk size 10485760
2015-10-27 18:37:35.0321 INFO DefaultQuartzScheduler_Worker-5 com.netflix.priam.aws.S3FileSystem Uploading to whatever-development/backup/eu-west-1/whatever/56713727820156410577229101239000783354/201510222039/SST/system/compactions_in_progress/system-compactions_in_progress-jb-3442-Filter.db with chunk size 10485760

Thanks.

tulumvinh:
Hello,

Priam will delete the file on disk (under "backups" for incrementals and "snapshots" for daily backups) if the upload is successful.  Base on your snippet of the log, the upload is in-progress.

Regards

sciclon:
Thanks for the replay,

I can see in the source code, for example the way Priam the "snapshots" in 
priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java

    private void clearSnapshot(final String snapshotTag) throws Exception                                                                                                                                                                 
     {
         new RetryableCallable<Void>()
         {
             public Void retriableCall() throws Exception
             {
                 JMXNodeTool nodetool = JMXNodeTool.instance(config);
                 nodetool.clearSnapshot(snapshotTag);
                 return null;
             }
         }.call();
     }

I can see how priam uses nodetool to delete the snapshots after uploading them to S3, in the other hand I see the incremental files are uploaded to S3 but not deleted in file:
priam/src/main/java/com/netflix/priam/backup/IncrementalBackup.java

Line 61:
       logger.debug("Scanning for backup in: {}", dataDir.getAbsolutePath());
        for (File keyspaceDir : dataDir.listFiles())
        {
            if (keyspaceDir.isFile())
        			continue;
            for (File columnFamilyDir : keyspaceDir.listFiles())
            {
                File backupDir = new File(columnFamilyDir, "backups");
[...]

PR #438, opened 2016-02-04
Adding logging when backup incremental files are not deleted.

sciclon:


sciclon:
Hey guys,

I had opened an "issue" about this, it is not an issue but as I said it is very common to have running bot processes (Cassandra and Priam) with different UID/GID, as Cassandra create the files with 644 it could fail when Priam tries to delete the uploaded files. 
Issue related: https://github.com/Netflix/Priam/issues/431

Thanks!

PR #464, opened 2016-04-25
Support for ACL's in Non-Default VPC

BlueDragonX:
This adds support to Priam for updating ACL's (security groups) in non-default VPC's.

Before this patch Priam would call DescribeSecurityGroups using the groupName. This call only searches EC2 Classic or the default VPC.

This change looks up the VPC ID of the running instance on start. The AWSMembership class has been updated to call DescribeSecurityGroups the VPC ID in conjunction with the group name. Where applicable this method is used to determine the group ID.

This has been tested against a new multi-region cluster in our in-house testing environment.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/IConfiguration.java
@@ -276,6 +276,11 @@ public interface IConfiguration
     public String getACLGroupName();
 
     /**
+     * Get the VPC ID of the current instance.
+     */
+    public String getVpcId();
+
+    /**
      * @return true if incremental backups are enabled
      */
     boolean isIncrBackup();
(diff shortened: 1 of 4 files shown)


pdehlke:
+1

Or, actually: +about-a-million.

Hallelujah.

tulumvinh:
Hello,

Thanks for the pull request, will review soon.

Regards

tulumvinh:
Hello BlueDragonX,

Two quick questions:
1. Is this change backward compatible?  Specifically, will it work with non-vpc environments?
2. It seems you hardcoded endpoints within PriamConfiguraiton.java, true?  If so, was that a mistake?

Regards

BlueDragonX:
1. It works with the default VPC but it does not work with EC2 Classic.
2. I've hardcoded the EC2 metadata endpoints. They're already hardcoded in that file.

tulumvinh:
Hi BlueDragonX,

I have an implementation which dynamically determines the env (classic, non-default vpc).  Once I complete the testing, I'll add you as a reviewer in the pull request

Regards

PR #472, opened 2016-05-26
Better logging when TuneCassandra fails

pdehlke:


tulumvinh:
It would add value to log the full path where C* is failing to write to the yaml file.

PR #491, opened 2016-06-16
Added rpc_min_threads and rpc_max_threads config options

dolphy17:
Added rpc_min_threads and rpc_max_threads config options with default values, because Cassandra doesn't support rpc_server_type=hsha setting with rpc_max_threads=unlimited (default) from 2.0.12.
With this patch Priam will support latest Cassandra version on 2.0 branch = 2.0.17

PR #509, opened 2016-07-26
Added rpc_min_threads and rpc_max_threads config options

tieum:
Same PR than #491 but for 3.x branch, so that Priam can support latest Cassandra version on 2.1 branch = 2.1.15


dolphy17:
Cool. Go ahead.

PR #511, opened 2016-07-29
cross-account support for AWS compute

ipapapa:
This PR adds support for cross-account C* deployments in AWS EC2 leveraging AWS Role assumption and multiple credentials. It is by default disabled. For example, one use case would be EC2/VPC hybrid deployment. In order to support cross-account deployments, there are two new Fast Properties, .ec2.roleassumption.arn and .vpc.roleassumption.arn in which the Amazon Resource Name (ARN) is defined in order to establish the credentials for the cross-account access.


vinaykumarchella:
LGTM.

PR #512, opened 2016-07-30
Fixing initialization bug of InstanceDataRetriever

dolphy17:
Hi Vinh,

This one is fixing the chicken-egg problem in config initializer.
To be honest, I have no idea if it ever worked for somebody. setupEnvVars() requires to RAC to be set, otherwise it crashes.

Please review.

Regards,
Andor


tulumvinh:
See merge #514

PR #513, opened 2016-07-30
Added support to eu-central-1 region in S3

dolphy17:
Adds support of "eu-central-1" region to S3 file system. Lets Priam work in Frankfurt AWS region.


tulumvinh:
Dolphy17, would you create a PR to the 2.x branch?

I plan to eliminate the need to have both 3.x and 2.x. I want to only have 2.x (for Cassandra 2.x versions) but I need to implement fetching the Cassandra version first (once you know the version, you can execute the applicable code).

dolphy17:
Sure, I can do it. Would you also like me to add support for the other european region as well "eu-west-1"?
Also, I think I forgot to add unit tests, are there any present for the US regions?

PR #514, opened 2016-08-01
Fix PriamConfiguration init order.

BlueDragonX:
Launching Priam without the `EC2_REGION` value set causes a NullPointerException due to `RAC` not being set on line 325 in `setupEnvVars()`. This rearranges the order of statements in `initialize()` to ensure `RAC` is set before `setupEnvVars()` is called.


tulumvinh:
BlueDragonX, would you create a PR to the 2.x branch?

I plan to eliminate the need to have both 3.x and 2.x.  I want to only have 2.x (for Cassandra 2.x versions) but I need to implement fetching the Cassandra version first (once you know the version, you can execute the applicable code).

BlueDragonX:
Here you go: https://github.com/Netflix/Priam/pull/515

PR #522, opened 2016-08-19
Added CommitLog archiving properties file path config option

matepeter90:
Hello,

The Cassandra version we use looks for the commitlog_archiving.properties file in Cassandra's home directory, not in the $cass_home/conf/ folder, so I routed the variable to a config option.

Thanks for Priam,
Peter

### DIFF
--- a/priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
@@ -124,6 +124,7 @@ public class PriamConfiguration implements IConfiguration
     private static final String CONFIG_INTERNODE_COMPRESSION = PRIAM_PRE + ".internodeCompression";
 
     private static final String CONFIG_COMMITLOG_BKUP_ENABLED = PRIAM_PRE + ".clbackup.enabled";
+    private static final String CONFIG_COMMITLOG_PROPS_FILE = PRIAM_PRE + ".clbackup.propsfile";
     private static final String CONFIG_COMMITLOG_ARCHIVE_CMD = PRIAM_PRE + ".clbackup.archiveCmd";
     private static final String CONFIG_COMMITLOG_RESTORE_CMD = PRIAM_PRE + ".clbackup.restoreCmd";
     private static final String CONFIG_COMMITLOG_RESTORE_DIRS = PRIAM_PRE + ".clbackup.restoreDirs";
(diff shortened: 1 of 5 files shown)


dolphy17:
Hi Vinh,

This change is a must for DSE support, because the config files are stored at different location.

Thanks,
Andor

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -165,10 +165,11 @@ public boolean isValidBackupDir(File keyspaceDir, File columnFamilyDir, File bac
             return false;        	
         }
 
-        String columnFamilyName = columnFamilyDir.getName();
+        String dirName = columnFamilyDir.getName();
 
+        String columnFamilyName = dirName.split("-")[0];
```
tulumvinh: This assumes that directory for CF is a hyphen.  Even though the code is backward compatible, should the delimiter not be a configurable property?

matepeter90:
Sorry, I meant to merge only the first commit. I wasn't aware, if I push to the branch again it will be added to the existing Pull Request.

I move the second patch away, because it's just work in progress.

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
matepeter90: Sorry, I meant to merge only the first commit. I wasn't aware, if I push to the branch again it will be added to the existing Pull Request.

I move the second patch away, because it's just work in progress.

priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
```diff
@@ -207,6 +208,7 @@
     private final String DEFAULT_PARTITIONER = "org.apache.cassandra.dht.RandomPartitioner";
     public static final String DEFAULT_AUTHENTICATOR = "org.apache.cassandra.auth.AllowAllAuthenticator";
     public static final String DEFAULT_AUTHORIZER = "org.apache.cassandra.auth.AllowAllAuthorizer";
+    public static final String DEFAULT_COMMITLOG_PROPS_FILE = "/conf/commitlog_archiving.properties";
```
tulumvinh: The default should be prefix with onfig.getCassHome() for backward compatibility.

matepeter90: It's prefixed runtime with getCassHome() when anything gets the file's location like it was before in the StandardTuner class.
I got the idea from getYamlLocation() function, to do it like this, only modification was that I moved the string part into a static default String.
Of course I change the patch, if you have any advice, I just didn't want to set the value of a default string in the constructor, because getCassHome is not a static function.

tulumvinh: Hi Peter, let me review your comments/pull request again.  Thanks

dolphy17:
Just a gentle nudge @tulumvinh to review the changes and merge if everything looks good.
Thanks, Andor

PR #527, opened 2016-10-03
Allow S3 Role assumption when interacting with S3 for backups

arunagrawal84:
- This will allow S3 Role assumptions when interacting with S3. 
- Code cleanup: Make IS3Credential an extension to ICredential.


vinaykumarchella:
LGTM

PR #554, opened 2017-03-08
Support for backup cron expression for snapshots via configuration 

arunagrawal84:
Support for backup cron expression for snapshots via configuration - "priam.backup.cron". This configuration expects a valid CRON expression to be set for it to take effect.

This takes precedence over "priam.backup.hour", so if you have overrides for "priam.backup.hour" they will be overridden with "priam.backup.cron" (if supplied). The default value of "priam.backup.cron" is set to empty CRON to provide backward compatibility. To disable the backups via "priam.backup.cron" use value of "0 0 5 31 2 ?" (Feb 31). For information on how to set up CRON timer refer to following documents -
 * @link http://www.quartz-scheduler.org/documentation/quartz-2.x/tutorials/crontrigger.html
 * @link http://www.cronmaker.com  To build new cron timer


priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -151,8 +152,8 @@ else if (UpdateSecuritySettings.firstTimeUpdated)
         scheduler.addTaskWithDelay(CassandraMonitor.JOBNAME,CassandraMonitor.class, CassandraMonitor.getTimer(), CASSANDRA_MONITORING_INITIAL_DELAY);
         
         // Start the snapshot backup schedule - Always run this. (If you want to
-        // set it off, set backup hour to -1)
-        if (config.getBackupHour() >= 0 && (CollectionUtils.isEmpty(config.getBackupRacs()) || config.getBackupRacs().contains(config.getRac())))
+        // set it off, set backup hour to -1) or set backup cron to "0 0 5 31 2 ?" (Feb 31)
```
vinaykumarchella: Instead of having it scheduled for Feb 31, not scheduling with -1 is a good idea

priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -151,8 +152,8 @@ else if (UpdateSecuritySettings.firstTimeUpdated)
...
+        if ((!StringUtils.isEmpty(config.getBackupCronExpression()) || config.getBackupHour() >= 0) && (CollectionUtils.isEmpty(config.getBackupRacs()) || config.getBackupRacs().contains(config.getRac())))
```
vinaykumarchella: You might also want to look for valid CronExpression.

priam/src/main/java/com/netflix/priam/scheduler/SchedulerType.java
```diff
@@ -0,0 +1,31 @@
+package com.netflix.priam.scheduler;
+
+/**
+ * Created by aagrawal on 3/8/17.
+ */
+public enum SchedulerType {
+    TIME("TIME"),CRON("CRON");
```
vinaykumarchella: Did you mean HOUR based instead of TIME ?

priam/src/main/java/com/netflix/priam/scheduler/SchedulerType.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    //Helper method to find the scheduler type - case insensitive as user may
+    //put value which are not right case.
+    public static SchedulerType valueOfIgnoreCase(String schedulerType) {
+        return SchedulerType.valueOf(schedulerType.toUpperCase());
+    }
+
+    public static SchedulerType lookup(String schedulerType) {
+        return valueOfIgnoreCase(schedulerType);
```
vinaykumarchella: This might throw a runtime exception if the input is not in one of the provided values. Something like below would be safer instead of throwing exceptions in the main code path. Throwing exception might work in current scenario but might for any other future users of this enum.

```
public static Blah fromString(String text) {
    for (Blah b : Blah.values()) {
      if (b.text.equalsIgnoreCase(text)) {
        return b;
      }
    }
    return null;
  }
```

priam/src/test/java/com/netflix/priam/FakeConfiguration.java
```diff
@@ -163,6 +164,16 @@ public int getBackupHour()
     }
 
     @Override
+    public String getBackupCronExpression() {
+        return null;
+    }
+
+    @Override
+    public SchedulerType getBackupSchedulerType() {
+        return SchedulerType.TIME;
```
vinaykumarchella: IMO, HOUR and CRON might make more sense than TIME and CRON.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -314,45 +297,63 @@ public Void retriableCall() throws Exception
...
+                if (hour >= 0) {
+                    cronTimer = new CronTimer(JOBNAME, hour, 1, 0);
+                    logger.info("Starting snapshot backup with backup hour: " + hour);
+                }else
+                    logger.info("Skipping snapshot backup as backup hour is less than 0: " + hour);
+                break;
+            case CRON:
+                String cronExpression = config.getBackupCronExpression();
+
+                if(!StringUtils.isEmpty(cronExpression) && cronExpression.equalsIgnoreCase("-1")){
```
vinaykumarchella: Should we be even in this stage if `cronExpression` is set to -1?

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -314,45 +297,63 @@ public Void retriableCall() throws Exception
...
+                    logger.info("Skipping snapshot backup as backup cron is set to NA");
+                }else
+                {
+                    if(StringUtils.isEmpty(cronExpression) || !CronExpression.isValidExpression(cronExpression))
+                        throw new Exception("Invalid CRON expression: " + cronExpression +
+                                ". Please use NA if you wish to disable backup else fix the CRON expression and try again!");
```
vinaykumarchella: Will it be `NA` or `-1`?

priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -55,91 +53,111 @@
...
+
             scheduler.addTask(UpdateSecuritySettings.JOBNAME, UpdateSecuritySettings.class, UpdateSecuritySettings.getTimer(id));
         }
 
         // Run the task to tune Cassandra
         scheduler.runTaskNow(TuneCassandra.class);
 
+        // Start the snapshot backup schedule - Always run this. (If you want to
+        // set it off, set backup hour to -1) or set backup cron to "NA"
+        if (SnapshotBackup.getTimer(config) != null && (CollectionUtils.isEmpty(config.getBackupRacs()) || config.getBackupRacs().contains(config.getRac()))) {
```
vinaykumarchella: `NA` or `-1`?

arunagrawal84: -1. Will update the comment.

priam/src/main/java/com/netflix/priam/scheduler/SchedulerType.java
arunagrawal84: I intentionally chose TIME as we can use schedulerType with CRON implementation for flush too which currently supports hour and daily.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
arunagrawal84: -1 again. will update.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
arunagrawal84: Yes, this is the only place where we check the value of cron expression.

priam/src/main/java/com/netflix/priam/scheduler/SchedulerType.java
arunagrawal84: Agreed. But I would like to check schedulerType and immediately throw new Exception("Unsupported SchedulerType") message instead.

vinaykumarchella: As mentioned above, `Throwing exception might work in current scenario but might for any other future users of this enum`. You might want to check for `null` when calling this method and act accordingly (In some cases it might be a default value and in other cases, it might be throwing an exception. Adding to the current behavior, when it is fail-fast exception, it is always good idea to show the path forward in exception messages. For .e.g., Instead of throwing a message `Unsupported SchedulerType`, something like below would be helpful for OSS community.

> Unable to start Priam as `blah` is not a supported SchedulerType, supported ScheduleTypes are `HOUR` or `CRON` formats.

priam/src/main/java/com/netflix/priam/scheduler/SchedulerType.java
vinaykumarchella: I think even flush implementation can be migrated to this format. `TIME` does not ring a bell that the values are expected in `HOUR` format. Even if we keep Flush implementation as is, these 2 formats are actually in sync - For e.g., Flush format  - `hour=0 or daily=10` which is in line with `SchedulerType=HOUR and backupHour =10.`

priam/src/test/java/com/netflix/priam/scheduler/TestSchedulerType.groovy
```diff
@@ -0,0 +1,52 @@
+package com.netflix.priam.scheduler
```
vinaykumarchella: Great.

build.gradle
```diff
@@ -57,6 +57,8 @@ subprojects {
       compile 'com.google.http-client:google-http-client-jackson2:1.19.0'
       provided 'javax.servlet:servlet-api:2.5'
       testCompile 'org.jmockit:jmockit:1.19'
+      testCompile "org.spockframework:spock-core:1.1-groovy-2.4-rc-2"
       testCompile 'junit:junit:4.8'
+      //testCompile 'org.assertj:assertj-core:3.6.2'
```
vinaykumarchella: Can you remove commented deps

PR #569, opened 2017-07-13
BackupStatusMgr refactor

arunagrawal84:
1. Make backup/snapshot status manager store additional details like meta file location. 
2. Ensure it can save last 60 snapshot days. 
3. Allow bindable values on how to save snapshot status in file/datastore. Default implementation stores in a file on the local instance. 
4. Non-backward compatible. The previous backup file gets overwritten with new values(null) during the upgrade process.


priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
```diff
@@ -15,251 +15,151 @@
...
 	  * Note:  A linkedlist was chosen for fastest removal of the oldest backup.
 	  */
-	private final LinkedList<BackupMetadata> bkups = new LinkedList<BackupMetadata>();
-	private final IConfiguration config;
-	private int capacity;
-	@Inject
-	public BackupStatusMgr(IConfiguration config) {
-		this.config = config;
-		this.capacity = 60;  //TODO: fetch from properties
+	Map<String, LinkedList<BackupMetadata>> backupMetadataMap;
```
vinaykumarchella: When using maps, can you put in comments what is it "keyed" by?

PR #576, opened 2017-07-19
Remove guava dependency in NewTokenRetriever.

arunagrawal84:


priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
```diff
@@ -67,7 +67,7 @@
     private final Predicate<PriamInstance> differentHostPredicate = new Predicate<PriamInstance>() {
         @Override
         public boolean apply(PriamInstance instance) {
-            return (!instance.getInstanceId().equalsIgnoreCase(DUMMY_INSTANCE_ID) && !instance.getHostName().equals(myInstance.getHostName()));
+            return (!instance.getInstanceId().equalsIgnoreCase(DUMMY_INSTANCE_ID));
```
vinaykumarchella: Can you provide comments here on what the algorithm is.

PR #578, opened 2017-08-08
Allow configuration of the size of large partitions that get logged by cassandra during compaction.

arunagrawal84:
Allow configuration of the size of large partitions that get logged by Cassandra during compaction.


priam/src/main/java/com/netflix/priam/IConfiguration.java
```diff
@@ -497,6 +497,12 @@
 
     public int getIndexInterval();
 
+    /*
+     * @return the warning threshold in MB's for large partitions encountered during compaction.
+     * Default value of 100 is used (default from cassandra.yaml)
+     */
+    public int getCompactionLargePartitionWarnThreshold();
```
vinaykumarchella: Can you suffix it with MB

PR #580, opened 2017-08-15
Code Refactoring

arunagrawal84:


priam/src/main/java/com/netflix/priam/notification/BackupEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package com.netflix.priam.notification;
+
+import com.netflix.priam.backup.AbstractBackupPath;
+
+/**
+ * Created by aagrawal on 8/11/17.
+ */
+public class BackupEvent {
```
vinaykumarchella: Description of the class would be helpful

priam/src/main/java/com/netflix/priam/notification/EventGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.netflix.priam.notification;
+
+/**
+ * Created by aagrawal on 8/11/17.
+ */
```
vinaykumarchella: Description of this entity would be helpful

PR #586, opened 2017-08-23
Moving all tuners to "optionTuners" package. 

arunagrawal84:
Code refactor.


priam/src/main/java/com/netflix/priam/tuner/GCTuner.java
```diff
@@ -0,0 +1,105 @@
...
+//            "-XX:ParallelGCThreads",
+//            "-XX:ConcGCThreads",
+//            "-XX:+ParallelRefProcEnabled",
+//            "-XX:+AlwaysPreTouch",
+//            "-XX:+UseTLAB",
+//            "-XX:+ResizeTLAB",
+//            "-XX:-UseBiasedLocking"
+//    ));
+
+    private static final Set<String> cmsOptions = new HashSet<>(Arrays.asList(
```
vinaykumarchella: Won't we have any thing in common?

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+        Files.write(new File(outputFile).toPath(), configuredJVMOptions);
+    }
+
+    /**
+     * Update the JVM options file for cassandra by updating/removing JVM options {@link IConfiguration#getJVMExcludeSet()} && {@link IConfiguration#getJVMUpsertSet()}, configuring GC {@link IConfiguration#getGCType()}etc.
+     *
+     * @return List of Configuration as String after reading the configuration from jvm.options
+     * @throws Exception when encountered with invalid configured GC type. {@link IConfiguration#getGCType()}
+     */
+    protected List<String> updateJVMOptions() throws Exception {
```
vinaykumarchella: How is it different from the above function? If it is different, can you come up with more descriptive name

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+        File jvmOptionsFile = new File(config.getJVMOptionsFileLocation());
+        validate(jvmOptionsFile);
+        final GCType configuredGC = config.getGCType();
+
+        //Read file line by line.
```
vinaykumarchella: This is not appropriate comment here.

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+        //Make a copy of upsertSet, so we can delete the entries as we process them.
+        Map<String, JVMOption> upsertSet = config.getJVMUpsertSet();
+
+        //Don't use streams for processing as upsertSet jvm options needs to be removed if we find them
+        //already in jvm.options file.
+        List<String> optionsFromFile = Files.lines(jvmOptionsFile.toPath()).collect(Collectors.toList());
+        List<String> configuredOptions = new LinkedList<>();
+        for (String line: optionsFromFile)
+        {
+            configuredOptions.add(updateConfigurationValue(line, configuredGC, upsertSet, excludeSet));
```
vinaykumarchella: updateConfigurationValue returns null in some cases, why is null check missing?

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private String updateConfigurationValue(final String line, GCType configuredGC, Map<String, JVMOption> upsertSet, Map<String, JVMOption> excludeSet) {
+
+        JVMOption option = JVMOption.parse(line);
+        if (option == null)
+            return line;
+
+        //Is parameter for GC.
+        GCType gcType = GCTuner.isOptionAvailable(option);
```
vinaykumarchella: Instead of `isOptionAvailable`,  `getGcType` makes more sense

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+        if (gcType != null) {
+            option.setCommented(gcType != configuredGC);
+        }
+
+        //See if option is in upsert list.
+        if (upsertSet != null && upsertSet.containsKey(option.getJvmOption())) {
```
vinaykumarchella: You might want to ensure that the casing is same. Otherwise, `containsKeys` might not work

priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
```diff
@@ -1,53 +1,49 @@
 /**
```
vinaykumarchella: Can you avoid clubbing formatting changes with the code changes? This helps in quality code review.

priam/src/main/java/com/netflix/priam/tuner/TuneCassandra.java
```diff
@@ -1,69 +1,68 @@
-/**
- * Copyright 2013 Netflix, Inc.
+/*
+ * Copyright 2016 Netflix, Inc.
```
vinaykumarchella: You might not want to update the Year in license

priam/src/main/java/com/netflix/priam/tuner/GCTuner.java
arunagrawal84: We would, but the idea is not to touch anything in common. Configuring GC will only enable/disable the parameters which are specific to GC.

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
arunagrawal84: It was intentional as JVM parameters are case sensitive AFAIK. My intention was to do comparison which cares about the case of JVM property otherwise I won't know which JVM option name I should pick (one provided in Cassandra jvm.options file or one provided by the user).

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
arunagrawal84: +1

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
arunagrawal84: updateConfigurationValue always return a String. If it is not JVM property it returns the "comment" as is. If a real JVM option then a JVM representation of that.

priam/src/main/java/com/netflix/priam/tuner/JVMOptionsTuner.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Update the JVM options file for cassandra by updating/removing JVM options {@link IConfiguration#getJVMExcludeSet()} && {@link IConfiguration#getJVMUpsertSet()}, configuring GC {@link IConfiguration#getGCType()}etc.
+     *
+     * @param outputFile File name with which this configured JVM options should be written.
+     * @throws Exception when encountered with invalid configured GC type. {@link IConfiguration#getGCType()}
+     */
+    public void updateJVMOptions(final String outputFile) throws Exception {
+        List<String> configuredJVMOptions = updateJVMOptions();
+
+        if (logger.isInfoEnabled())
```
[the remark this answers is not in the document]
arunagrawal84: We are using logger.info level to log. This is to check if INFO is actually ON.

priam/src/main/java/com/netflix/priam/IConfiguration.java
```diff
@@ -39,6 +42,26 @@
...
 
     /**
+     * @return Path to jvm.options file. This is used to pass JVM options to Cassandra.
+     */
+    public String getJVMOptionsFileLocation();
+
+    /**
+     * @return Type of garbage collection mechanism to use for Cassandra. Supported values are CMS,G1GC
+     */
+    public GCType getGCType() throws UnsupportedTypeException;
```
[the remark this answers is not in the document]
arunagrawal84: There are multiple places where we are sending complex types. Would be better to club that with the move to Archaius2 and bring validation module for configuration (as startup requirement).

PR #612, opened 2017-10-19
Switch to Oracle JDK 8

dtrebbien:
Per a comment on travis-ci/travis-ci#7884 Oracle JDK 7 is no longer supported by Travis because the installer is now only available to those with an Oracle Support account.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-- oraclejdk7
+  - oraclejdk8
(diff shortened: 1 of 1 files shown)


dtrebbien:
From the [build output](https://travis-ci.org/Netflix/Priam/jobs/289771306), it appears that there is some issue running gradlew on OpenJDK 7. Others have experienced an issue as well (e.g. docker-library/openjdk#117)

PR #613, opened 2017-10-19
SLF4J-related fixes against the 3.x branch

dtrebbien:


dtrebbien:
Note that there were a couple of cases where String format conversion `%S` was used. This is like `%s` except that the interpolated string is converted to upper case. It appears, however, that the intent was for `%s` to be used.

PR #622, opened 2017-11-07
SNS Notification

arunagrawal84:
* New feature: SNS Notifications on backup events
* Bug fix: disable incrementals if snapshot is disabled


priam/src/main/java/com/netflix/priam/notification/AWSSnsNotificationService.java
```diff
@@ -0,0 +1,121 @@
...
+public class AWSSnsNotificationService implements INotificationService {
+	private static final Logger logger = LoggerFactory.getLogger(AWSSnsNotificationService.class);
+
+	private IConfiguration configuration;
+	private AmazonSNS snsClient;
+	private IMetricPublisher metricPublisher;
+	private IMeasurement<Object> notificationMeasurement;
+	
+	@Inject
+	public AWSSnsNotificationService(IConfiguration config, IAMCredential iamCredential
```
jolynch: Since this is the default implementation now, do we have a good Noop for IAMCredentials? Looks like metricspublisher has a noop and config is required.

arunagrawal84: IAM Credentials are used for every AWS operations and thus is always provided for Priam to work

PR #623, opened 2017-11-08
3.11vnodesupport

zmarois:


zmarois:
A few overarching changes required:

* Make startup not require a token. When using vnodes, we don't need to generate a token
* Separate the notion of the initial token from the backup identifier for the node. When using vnodes, we still need a unique identifier for the backup, but all 256 initial tokens seems too long, and we still need to set initial token at startup when loading from a backup, so we can't overload token as backup id because both are used for vnodes.
* Save the instance's tokens in the snapshot in order to start the node loading it with the initial_token that was used when taking the snapshot, [step 5 in this overview](https://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsSnapshotRestoreNewCluster.html?hl=restore%2Csnapshot)

arunagrawal84:
@zmarois sorry I have not looked yet into your pull request. I am on vacation. I have added reviewers to your pull request for time being. 
Also, to answer your question, yes please submit a pull request for 3.x as well because we are trying to keep 3.x and 3.11 in feature parity here. Reason being 3.x is probably most widely used branch. `master` at this point is too far from 3.x and we plan to merge 3.x into master but it has been delayed due to new features/improvements.

zmarois:
No worries @arunagrawal84. I've got a fork I'm building and integration testing with; I just found a problem today that I have since patched in, so I'm not 100% sure that this is done. I appreciate any feedback the reviewers you added have, and I'll comment when I have integration tested this better, including restoring from backups.

zmarois:
I've now tested backups and restores using this change and have confirmed the data loads correctly. However, I also tried a restore without the complexity of setting the initial_token to be the same as when the backup was taken, per [step 5 in this overview](https://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsSnapshotRestoreNewCluster.html?hl=restore%2Csnapshot), and saw the data still load correctly. Perhaps my comparison of the data performed a read-repair of all the data, and that resolved data being initially loaded onto the wrong nodes. Further, I'm not quite sure how after setting initial_token on restore I'd then be able to add nodes that are vnodes, so I'd like to remove the complexity of setting initial_token when restoring from a backup at all, or at least make it configurable to do so.

I'm going to strip that complexity, then file the PR against the 3.x branch. 

One more thing to note: I didn't do my testing against this branch itself, but a fork with a few more changes/features to make this easier to install into my current infrastructure, namely:
* The ability to run outside of an Auto-scaling group (mostly [committed in this](https://github.com/zmarois/Priam/commit/6dfd180a0408f6f99e6ecd91cf92c37b53d87acd). The ASG is used for a few purposes, which I needed to provide an alternative implementation for:
  * To calculate balanced token ranges from the total cluster size when not using vnodes. As I am using vnodes, I chose to limit this change to only being supported for vnode instances, but could be implemented with an instance name pattern.
  * To determine if a node registered in SDB's InstanceIdentity domain is still living or not, which I just implemented as "is this node in the list of all instances returned by ec2 client's describeInstances"
  * To wait for all nodes (thus all nodes need to be known, which the ASG provides) to be registered with SDB before returning seeds when running in a single-zone deployment. I have to admit, I don't understand the reasoning for this edge case. I could implement this again with an EC2 instance name pattern, but as I'm not testing with a single AZ, I just left this case as not implemented under the vnode feature. Further, a large benefit of vnodes is the simpler flexibility in changing the size of the cluster, so the notion of "are all up" is less applicable.
  * To determine the app id to query SDB with for configuration, prior to the configuration which is in SDB specifying the cluster name. This was already inferred from system property ASG_NAME if it was set, so I just used that.
* The ability to run nodes on a private network in a private VPC, related to the [conversation here](https://github.com/Netflix/Priam/pull/624#pullrequestreview-77487278), most of the [change to this file in this commit](https://github.com/zmarois/Priam/commit/6dfd180a0408f6f99e6ecd91cf92c37b53d87acd#diff-eba8ef5ed57356f9a548be0a4c91ab72), although I prefer the deferring/encapsulation pattern in that PR with the conversation.
* A static seed retriever, implemented [in this commit](https://github.com/zmarois/Priam/commit/c5028f588139a533caa3ce0589267f0caf3572b0), but interface for it introduced [here](https://github.com/zmarois/Priam/commit/6dfd180a0408f6f99e6ecd91cf92c37b53d87acd#diff-57e65d9b995095b40687f4998fde2e2d). I do not dislike Priam's mechanism to provide seeds, but wanted to reduce the differences to my current infrastructure (which uses a SimpleSeedProvider) for fewer moving parts for now. I chose this pattern, as opposed to just using the built in SimpleSeedProvider over NFSeedProvider, so the PriamStartupAgent still waits for Priam to run before starting.
* The ability to run both SDB domains in the same region, [this commit](https://github.com/zmarois/Priam/commit/5921c6dbbcf05047a072e9f2dbd9bf0e01306819)
* The ability to configure credentials, membership provider, and seed retriever at runtime instead of compile time, just so I didn't need to build different artifacts to vary these, [this commit](https://github.com/zmarois/Priam/commit/f62650e8759f590495088c4b251c153d3a60a8f2)
* Auth on the jmx port, [this commit](https://github.com/zmarois/Priam/commit/5d092138bf7be75e70e33f1e80bf19a1c5e110c8). I probably should make the CassandraTuner write this out instead of relying on the existing config to have it. 
* Ability to override some configuration from system properties that are in the properties config file by inverting the precedence, [this commit](https://github.com/zmarois/Priam/commit/55ea64b00f39bc740e8be8b951b5ad62e49512f6)
* A check when replacing a dead node that some node in the existing cluster actually thinks its dead, as a node that failed to ever really join the cluster could be in SDB, but the replacement node can't actually replace its place in the cluster, [this commit](https://github.com/zmarois/Priam/commit/c8ad578077953d7c028e59e3bf34f75bc91c5dbe). Really just helpful during testing when I was having nodes not come up due to configuration issues.
* Actually setting listen_address to something other than null, [this commit](https://github.com/zmarois/Priam/commit/bda52704ef5fde93272aa6866dbddd2e9b01e356). I have to admit, [per the docs on this property](https://docs.datastax.com/en/cassandra/3.0/cassandra/configuration/configCassandra_yaml.html?hl=listen_address#configCassandra_yaml__listen_address), I do not understand how this works being hard-coded to null for anything but single-node installation, unless listen_interface is set in the config externally
* Letting backupStatusFileLocation be variable, [this commit](https://github.com/zmarois/Priam/commit/55d7a6c6115bc14e9c85b61dfad3cf3ccc534089)

If you'd consider any of these features useful, I'd gladly file separate feature PRs for them.

costimuraru:
@zmarois nice work! I look forward to seeing this PR merged, so that we can benefit from VNodes.

priam-cass-extensions/src/main/java/com/netflix/priam/cassandra/extensions/PriamStartupAgent.java
```diff
@@ -44,14 +44,18 @@ private void setPriamProperties()
...
         String replacedIp = "";
         String extraEnvParams = null;
         
         while (true)
         {
             try
             {
-                token = DataFetcher.fetchData("http://127.0.0.1:8080/Priam/REST/v1/cassconfig/get_token");
+                isExternallyDefinedToken = Boolean.parseBoolean(DataFetcher.fetchData("http://127.0.0.1:8080/Priam/REST/v1/cassconfig/is_externally_defined_token"));
+                if (isExternallyDefinedToken) {
```
zmarois: I needed to allow token to not be set, but still distinguish between Priam being fully initialized but not having a token and it not running at all. 

I could have instead differentiated these two cases with a 500 when it hasn't fully initialized, and make this startup agent treat 500s different than a 204.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
```diff
@@ -104,7 +107,7 @@ public InstanceIdentity(IPriamInstanceFactory factory, IMembership membership, I
         init();
     }
 
-    public PriamInstance getInstance() {
+    PriamInstance getInstance() {
```
zmarois: Token was previously used both as the initial_token and as the identifier of a node's backup in the context of the entire cluster. 

When using vnodes, I don't need an initial token, but I do still need an identifier for the backup, so I wanted to separate these two concerns. Because the backup identifier is redundant with data already in SDB, it felt better not to put in in the PriamInstance class, and instead use PriamInstance purely as the Data Access DTO, but InstanceIdentity as the business model.

priam/src/main/java/com/netflix/priam/resources/RestoreServlet.java
```diff
@@ -160,25 +164,12 @@ private void restore(String token, String region, Date startTime, Date endTime,
...
+            priamServer.getId().setBackupIdentifier(origBackupIdentifier);
         }
         tuner.updateAutoBootstrap(config.getYamlLocation(), false);
         cassProcess.start(true);
     }
 
-    /**
-     * Find closest token in the specified region
-     */
-    private String closestToken(String token, String region) {
```
zmarois: I moved this into RestoreTokenSelector to keep the logic for determining the closest token in one place

priam/src/main/java/com/netflix/priam/restore/RestoreTokenSelector.java
```diff
@@ -54,11 +62,52 @@ public RestoreTokenSelector(ITokenManager tokenManager, @Named("backup") IBackup
      *            Date for which the backups are available
      * @return Token as BigInteger
      */
-    public BigInteger getClosestToken(BigInteger tokenToSearch, Date startDate) {
+    public BigInteger getClosestToken(String tokenToSearch, Date startDate)
```
zmarois: The changes to this are purely to improve the error message when finding the closest token is invalid, because the snapshot was taken with multiple tokens.

priam/src/test/java/com/netflix/priam/identity/DoubleRingTest.java
```diff
@@ -15,19 +15,13 @@
  *
  */
 
-package com.netflix.priam.backup.identity;
+package com.netflix.priam.identity;
```
zmarois: I moved these tests to same package as the classes they test, so they have access to the now-package-private instance properties.

arunagrawal84:
@zmarois wanted to give more context: This requires some changes in our infrastructure to accommodate vnodes which is why it is taking longer than expected.

zmarois:
@arunagrawal84 no worries. I appreciate the check-in. I tried to write this in a way in which, if you still have num_tokens set to 1 (which is still the default), Priam should operate identical to how it did without this branch.

If you have concrete feedback on something I did here that makes Priam not backwards-compatible, let me know; I'd gladly fix it. If its just something internal, perhaps your own fork that I do not realize isn't compatible with this, I understand.

darkpssngr:
Any updates on this ?

zmarois:
FWIW, we have been running a build of [our fork with this](https://github.com/Cimpress-MCP/Priam) in our production environment for about two months. @darkpssngr feel free to try out a build of our 3.11 branch. 

Note it does have a few more changes in it, namely:

* [A static seed definition](https://github.com/Netflix/Priam/commit/b64dd32476905ad8ff82efe206872ca89a7bea13). I do not dislike Priam's mechanism to provide seeds, but this was useful to both add nodes running Priam to an existing cluster that is not.
* [A check when replacing a dead node that some node in the existing cluster actually thinks its dead](https://github.com/Netflix/Priam/commit/7b8d8be78edb4b2476d572e2341630961a6ba05c), as a node that failed to ever really join the cluster could be in SDB, but the replacement node can't actually replace its place in the cluster. Really just helpful during testing when I was having nodes not come up due to configuration issues.
* [Ability to override some configuration from system properties that are in the properties config file by inverting the precedence](https://github.com/Netflix/Priam/commit/39775067f395069c181b4a25d222cec1baa8c970)
* [The ability to configure credentials at runtime instead of compile time](https://github.com/Netflix/Priam/commit/010c5d19f044f2a195604177959f9308acfcf922), just so I didn't need to build different artifacts to vary these.
* [The ability to run outside of an Auto-scaling group](https://github.com/Netflix/Priam/commit/640c7c086b0e2217bba3563deafb7cf08bf4f892). Only possible with vnodes because we don't need to calculate token ranges (based on total cluster size defined by ASG)
* [Actually setting listen_address to something other than null](https://github.com/Netflix/Priam/commit/ed98b761c677fab26bf32ee568f18efdab003e93). I have to admit, [per the docs on this property](https://docs.datastax.com/en/cassandra/3.0/cassandra/configuration/configCassandra_yaml.html?hl=listen_address#configCassandra_yaml__listen_address), I do not understand how this works being hard-coded to null for anything but single-node installation, unless listen_interface is set in the config externally
* [Not validating md5 in multipart upload](https://github.com/Netflix/Priam/commit/62fb2222140f45e707faed0b61cf26c06fc4dc91), which doesn't work for encrypted S3 buckets, which I still want to [push back](https://github.com/Netflix/Priam/issues/655) into the Netflix repo when I get a chance.

Other than setting listen_address (which I don't understand) I don't think any of these would be breaking from the base Netflix repo if you don't explicitly opt into them, so you should be able to get back to base easy enough.

darkpssngr:
@zmarois Cool thanks . I'll try it out. Ours is a new cluster. So, I dont think backward compatibility is an issue here 👍

zmarois:
Fair. Or consider it forwards compatibility with the base Netflix branch if you switch back to it.

sbilello:
@zmarois and @darkpssngr  Do you currently use this: https://github.com/Cimpress-MCP/Priam ?

zmarois:
@sbilello unfortunately, no. My team no longer uses that. We did use it for about 6 months and it did enable us to perform a restore of a lost node at a particularly troublesome time. However, we have since moved off of Cassandra; we just couldn't justify the operational cost of managing it compared to using a hosted solution (FWIW, we actually ended up going with a hosted Postgres database, which also trades off some performance against the costs of hosted Cassandra).

If you would like to pick it up, I'd gladly fill you in with whatever knowledge I have of it. My largest open concern with it regarded assigning the correct token ranges on restore,  per [step 5 in this overview](https://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsSnapshotRestoreNewCluster.html?hl=restore%2Csnapshot). While on a full restore I had seen all data seem to propagate, it did balance it out on restoring (which just takes too long), so I think something needs to be done to pre-assign the token ranges such that nodes start up and take the correct token ranges for the restored data.

I believe @darkpssngr , per [this comment](https://github.com/Netflix/Priam/issues/649#issuecomment-355801520) opted to utilize ebs snapshots. I dunno if that has worked well for them or if they have tested a partial or full restore with it.

trashguy:
Any movement on this?

arunagrawal84:
vnodes are more complicated w.r.t. backups and restores. The token maps which an instance holds can change considerably between a backup and its subsequent incremental backup. Restoring a cluster with vnodes is another challenge. 
We want to pursue vnodes only when we have a complete solution for backups and restores, which produces consistent results.

PR #624, opened 2017-11-16
Set broadcast_rpc_address to private ip for COPY functionality to work.

arunagrawal84:


priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
```diff
@@ -275,21 +263,22 @@ public PriamConfiguration(ICredential provider, IConfigSource config, InstanceEn
...
+                this.instanceDataRetriever =  (InstanceDataRetriever) Class.forName("com.netflix.priam.identity.config.AWSVpcInstanceDataRetriever").newInstance();
+            } else {
+                throw new IllegalStateException("Unable to determine environemt (vpc, classic) for running instance.");
+            }
         } catch (Exception e) {
             throw new IllegalStateException("Exception when instantiating the instance data retriever.  Msg: " + e.getLocalizedMessage());
         }
 
         RAC = instanceDataRetriever.getRac();
-        PUBLIC_HOSTNAME = instanceDataRetriever.getPublicHostname();
```
zmarois: I was seeing a problem running in a VPC and have a [similar change in a fork](https://github.com/zmarois/Priam/commit/1f6e3a0064d9cf769f3a0fdd23cbedf4e029c133#diff-eba8ef5ed57356f9a548be0a4c91ab72) I am testing with. (Sorry, there are more changes in that commit to enable me to run outside of an ASG)

Is this why you deferred the evaluation of this?

Further, as far as I could tell, `/meta-data/public-ipv4` also does not return anything when running in a VPC, and I needed to defer the following line also.

arunagrawal84: @zmarois My intention was not to delay this but more to encapsulate the instance information at one place and left in half because of the time crunch, instead of really cleaning up everything. 
AFAIK, VPC has 2 flavors - VPC internal (only private-IP available) and VPC external (private and public-IP available). For Netflix, we always host Cassandra in VPC external as we have multi-region deployments where we need public-IP for cross-region communication. Thus, delaying that does not have any impact on our infrastructure.

zmarois: Ah, makes sense. We are keeping our instances private for now, which is causing this difference.

Thanks for filling me in.

PR #629, opened 2017-12-05
Autostart Cassandra from Priam

jolynch:
I took a whack at having Priam auto-remediate a dead Cassandra. By default the starts are limited to once per hour.

I'm not sure the best way to test this, I tried testing it with Mocks but the rate limiter makes it tough to reliably test (threads, time, etc ...). Do we have a better way to do functional tests?


jolynch:
3.11 PR @ #630

PR #630, opened 2017-12-05
Autostart cassandra 311

jolynch:
3.11 pull of https://github.com/Netflix/Priam/pull/629

I had to fixup some conflicts, but I think it's still good.


arunagrawal84:
LGTM

Issue #597, opened 2017-09-18
Priam is writing lots of debug logs to catalina.out


NickBlow:
We are trying to set up Priam using the 3.11 branch in Tomcat. It's all working fine, but we are having issues with logs. We're struggling to get Priam to pick up the log4j properties, and it's just logging huge amounts of debug info to catalina.out.

Our log4j config is the following (priam_log_file is injected via ansible). We are putting it in priam/src/main/resources/log4j.properties, then building with the provided gradle script. We also tried to put it into priam-web/src/main/resources/log4j.properties and had the same issue. I'm sure we're missing something obvious, but we're struggling to see it.

```log4j.debug=FALSE
log4j.rootLogger=ERROR, FILE, stdout
log4j.appender.FILE=org.apache.log4j.RollingFileAppender
log4j.appender.FILE.File={{priam_log_file}}
log4j.appender.FILE.MaxFileSize=5MB
log4j.appender.FILE.MaxBackupIndex=5
log4j.appender.FILE.layout=org.apache.log4j.PatternLayout
log4j.appender.FILE.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss.SSSS} %p %t %c %m%n
log4j.appender.stdout=org.apache.log4j.ConsoleAppender
log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
log4j.appender.stdout.layout.ConversionPattern=%5p %d{HH:mm:ss,SSS} %m%n
```

zmarois:
@NickBlow dunno if you figured out what was causing this, but I think it was multiple slf4j bindings on the classpath. I believe https://github.com/Netflix/Priam/pull/631 and https://github.com/Netflix/Priam/pull/632 fix this.

NickBlow:
@zmarois we ended up using filebeat to work around it rather than actually trying to dig deep into what was causing it. that sounds plausible. will close now and reopen should it not be fixed by those.

PR #632, opened 2017-12-07
excluding module with duplicate slf4j binding

zmarois:
I was seeing duplicate slf4j bindings in the logs, causing the log4j one not to be picked up sometimes:
```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/zmarois/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-log4j12/1.7.25/110cefe2df103412849d72ef7a67e4e91e4266b4/slf4j-log4j12-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/zmarois/.gradle/caches/modules-2/files-2.1/ch.qos.logback/logback-classic/1.1.2/b316e9737eea25e9ddd6d88eaeee76878045c6b2/logback-classic-1.1.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
```

I believe this is the cause for issue https://github.com/Netflix/Priam/issues/597

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,11 @@ allprojects {
     repositories {
         jcenter()
     }
+    
+    configurations {
+      all*.exclude group: 'ch.qos.logback', module: 'logback-classic'
+      all*.exclude group: 'ch.qos.logback', module: 'logback-core'
+    }
 
     dependencies {
       compile 'org.apache.commons:commons-lang3:3.5'


build.gradle
```diff
@@ -36,7 +36,9 @@ allprojects {
       compile 'com.googlecode.json-simple:json-simple:1.1.1'
       compile 'org.xerial.snappy:snappy-java:1.1.2.6'
       compile 'org.yaml:snakeyaml:1.18'
-      compile 'org.apache.cassandra:cassandra-all:2.1.17'
+      compile ('org.apache.cassandra:cassandra-all:2.1.17'){
```
zmarois: I could instead exclude this from any dependency

```
    configurations {
        all*.exclude group: 'ch.qos.logback', module: 'logback-classic'
    }
```

arunagrawal84: For reference, internally we always exclude the module for configurations.all: 
    exclude module: 'logback-classic' //by artifact name
    exclude module: 'logback-core' //by artifact name

The reason why we never experienced this. 
@zmarois your thoughts?

zmarois: I'm not quite following your suggestion. Do you suggest

1. Along my comment above, just add the exclude to configurations.all
2. Also exclude 'logback-core'
3. Just do the exclude in my internals and leave this repo without the exclude

I'm totally on board with 1 and/or 2. I have done 3 already, just considering there was an issue explicitly raised against this, seemed worthwhile to share it.

arunagrawal84: sorry, i was not clear. I am suggesting (1) and (2). I believe this would be a problem for others as well so better to push it in the trunk so we don't duplicate efforts.

zmarois: Gotcha. I'll push that change. Thanks for the review!

zmarois: Alright, I've excluded both logback-classic and logback-core. I don't have a ton of experience with gradle, so my syntax might not be as terse as it could be, but I confirmed the jars are excluded from the final build artifact.

PR #646, opened 2017-12-21
Running both sdb clients in the same region

zmarois:
Uses the same (and not-deprecated) client builder as the sdb client for PriamProperties, which makes them run in the same region

@arunagrawal84 Do you want this completely as-is, or do you want it to be configurable and default to us-east-1, per your docs?

I haven't found the AWS docs on how the two builders work in this regard, but my experience was that new AmazonSimpleDBClient creates a client for us-east-1, but AmazonSimpleDBClient.builder() creates a client in the region my instances are in (I am using IAMCredential).

I just worry about breaking existing clusters.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/IConfiguration.java
@@ -265,6 +265,11 @@ public interface IConfiguration {
     public String getRestoreSnapshot();
 
     /**
+     * @return Get the region to connect to SDB for instance identity
+     */
+    public String getSDBInstanceIdentityRegion();
+
+    /**
      * @return Get the Data Center name (or region for AWS)
      */
     public String getDC();
--- a/priam/src/main/java/com/netflix/priam/aws/SDBInstanceData.java
+++ b/priam/src/main/java/com/netflix/priam/aws/SDBInstanceData.java
@@ -17,6 +17,7 @@
 package com.netflix.priam.aws;
 
 import com.amazonaws.AmazonServiceException;
+import com.amazonaws.services.simpledb.AmazonSimpleDB;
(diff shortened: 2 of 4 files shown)


arunagrawal84:
You are correct in that assumption. Using builder will always ensure that it tries to connect in your "local" region, unless specified otherwise. We can force the region by [withRegion](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/client/builder/AwsClientBuilder.html#withRegion-java.lang.String-). 

That being said, I _think_ the old code should be connecting to "local" region as well as we do not specify any region to connect to. Can you point me where in the documentation it says we always connect to `us-east-1`? @vinaykumarchella your opinion on this?

zmarois:
Sorry, I dropped the link to your docs.

https://github.com/Netflix/Priam/wiki/Setup#s3-bucket-and-simpledb-domains

My experience was that the old code connected to us-east by default. I can double check that; Its been about a month since I've messed with the SDB clients.

zmarois:
With the old code, I get a AmazonSimpleDBClient that has endpoint "https://sdb.amazonaws.com", which is [us-east-1](http://docs.aws.amazon.com/general/latest/gr/rande.html#sdb_region).

With my new code, I get a AmazonSimpleDB that has an endpoint "https://sdb.eu-west-1.amazonaws.com" which is eu-west-1, but as you say, would be my local region (My nodes are in eu-west-1).

arunagrawal84:
I believe in that case right thing (for backward compatibility) would be default to `us-east-1` in the upstream branch even when I don't like cross-region traffic. 
You have to override that default in your configuration so you connect to the local region (the right thing to do). 
Also, thanks for catching this!

zmarois:
Alright, its configurable and defalted to us-east-1

PR #648, opened 2017-12-21
Move to latest gradle and netflixoss

arunagrawal84:


arunagrawal84:
This is to avoid issues which folks are having building master branch. Note that `master` branch where it stands is still considered `deprecated` from our view until and unless 3.x merges to `master`.

PR #665, opened 2018-03-05
Add graceful stopping option

jolynch:
Introduces a new configuration option gracefulDrainHealthWaitSeconds
which allows Priam controlled stops to fail healthchecks for a
configurable number of seconds before draining (via nodetool drain) and
stopping Cassandra (via the stop script). For backwards compatibility
this functionality is _disabled_ by default. If you want it change the
configuration to a positive number.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/IConfiguration.java
@@ -76,7 +76,7 @@ public interface IConfiguration {
     /**
      * @return int representing how many seconds Priam should fail healthchecks for before gracefully draining (nodetool drain)
      * cassandra prior to stop. If this number is negative then no draining occurs and Priam immediately stops Cassanddra
-     * using the provided stop script. If this number is >= 0 then Priam will fail healthchecks for this number of
+     * using the provided stop script. If this number is &gt;= 0 then Priam will fail healthchecks for this number of
      * seconds before gracefully draining cassandra (nodetool drain) and stopping cassandra with the stop script.
      */
     public int getGracefulDrainHealthWaitSeconds();
--- a/priam/src/main/java/com/netflix/priam/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/IConfiguration.java
@@ -74,6 +74,14 @@ public interface IConfiguration {
     public String getCassStopScript();
 
     /**
+     * @return int representing how many seconds Priam should fail healthchecks for before gracefully draining (nodetool drain)
+     * cassandra prior to stop. If this number is negative then no draining occurs and Priam immediately stops Cassanddra
+     * using the provided stop script. If this number is positive then Priam will fail healthchecks for this number of
+     * seconds before gracefully draining cassandra (nodetool drain) and stopping cassandra with the stop script.
+     */
+    public int getGracefulDrainHealthWaitSeconds();
+
+    /**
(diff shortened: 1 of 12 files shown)


priam/src/main/java/com/netflix/priam/IConfiguration.java
```diff
@@ -73,6 +73,14 @@
      */
     public String getCassStopScript();
 
+    /**
+     * @return int representing how many seconds Priam should fail healthchecks for before gracefully draining (nodetool drain)
+     * cassandra prior to stop. If this number is negative then no draining occurs and Priam immediately stops Cassanddra
```
vinaykumarchella: Based on the description of this variable, when it is -ve number, should we still drain and stop C* but do not fail any health check before draining?

priam/src/main/java/com/netflix/priam/defaultimpl/CassandraProcessManager.java
```diff
@@ -154,6 +157,34 @@ public void stop() throws IOException {
...
+                    logger.error("Exception draining Cassandra, could not drain. Proceeding with shutdown.", e);
+                }
+                // Once Cassandra is drained the thrift/native servers are shutdown and there is no need to wait to
+                // stop Cassandra. Just stop it now.
+            });
+
+            // In case drain hangs, timeout the future and continue stopping anyways.
+            try {
+                drainFuture.get(2 * config.getGracefulDrainHealthWaitSeconds(), TimeUnit.SECONDS);
+            } catch (ExecutionException | TimeoutException | InterruptedException e) {
```
vinaykumarchella: Do you want to use different error here, stating that waited config amount of time for the drain to complete, but proceeding? so that we can identify pattern if any

vinaykumarchella: And a different timeout config/ constant. So that we give enough time for drain always regardless of getGracefulDrainHealthWaitSeconds setting

priam/src/main/java/com/netflix/priam/defaultimpl/CassandraProcessManager.java
```diff
@@ -154,6 +157,34 @@ public void stop() throws IOException {
         stopCass.redirectErrorStream(true);
 
         instanceState.setShouldCassandraBeAlive(false);
+        if (config.getGracefulDrainHealthWaitSeconds() > 0) {
```
vinaykumarchella: Can we put a force option so that operators can call C* stop immediately without waiting for the drain in *some* scenarios?

priam/src/main/java/com/netflix/priam/IConfiguration.java
jolynch: I don't think so unless we want more configuration options. I think that a single second of wait time would not be a big deal for users who just want drain functionality. (So they'd set it to 1)

priam/src/main/java/com/netflix/priam/defaultimpl/CassandraProcessManager.java
jolynch: Yes, I can fix that up.

priam/src/main/java/com/netflix/priam/IConfiguration.java
jolynch: Oh I see, so 0+ would do the right thing and "-1" is disabled, sure.

Issue #668, opened 2018-03-16
Default path properties not matching documentation


Glamdring:
The documentation lists the default values for the path properties are /etc/cassandra ... https://github.com/Netflix/Priam/wiki/2.x

However, Priam.properties has them set to /mnt/cassandra and they get loaded. So, effectively, the documentation is wrong

Issue #667, opened 2018-03-16
Rename parent project


Glamdring:
Currently both the parent project and one of the subprojects are called "priam" (if we disregard the case). As a consequence, Eclipse fails to import them as they would result in a project with the same name. Maybe the parent project should be called "priam-parent"?

Issue #666, opened 2018-03-16
Option to customize SimpleDB domains


Glamdring:
Currently the domains are hardcoded ("PriamProperties" and "InstanceIdentity"). They should ideally be configurable, Currently, for example, a clodformation stack template doesn't allow you to specify exact names (it creates them with "stackname-PriamProperties-someSuffix", so Priam doesn't work.

PR #670, opened 2018-03-19
Fixing issues in deployment, including #666, #667 and #668

Glamdring:


priam/src/main/java/com/netflix/priam/aws/SDBInstanceData.java
```diff
@@ -45,17 +50,24 @@
...
     private final ICredential provider;
     private final IConfiguration configuration;
-
+    private final String domain;
+    
     @Inject
     public SDBInstanceData(ICredential provider, IConfiguration configuration) {
         this.provider = provider;
         this.configuration = configuration;
+        String configuredDomain = System.getProperty("priam.sdb.instanceidentity.domain");
```
arunagrawal84: Can we wrap this configuration into IConfiguration?

Glamdring: Sure; will do asap

PR #671, opened 2018-03-19
Compactions on CRON

arunagrawal84:
1. Bug fix: During restore do not start Cassandra, if `priam.cass.manual.start.enable` is set to true.

2. Compactions on schedule: Operator can set up compactions on CRON job using configurations: `priam.compaction.cron` and `priam.compaction.cf.filter`.


priam/src/main/java/com/netflix/priam/cluster/management/IClusterManagement.java
```diff
@@ -15,12 +15,72 @@
...
+        super(config);
+        this.taskType = taskType;
+        this.metricPublisher = metricPublisher;
+        this.measurement = measurement;
+    }
+
+    @Override
+    public void execute() throws JMXConnectionException, Exception {
+        JMXConnectorMgr connMgr = null;
+        List<T> result = null;
```
vinaykumarchella: Can you simplify and introduce a `Result` entity instead of List<T>, not always I might have List. `Result` entity could be a very simple interface and calling toString method on implementations of `Result` could be used in the output that you are trying to use.

priam/src/main/java/com/netflix/priam/cluster/management/IClusterManagement.java
```diff
@@ -15,12 +15,72 @@
...
+        try {
+            connMgr = new JMXConnectorMgr(config);
+        }catch (IOException | InterruptedException e){
+            measurement.incrementFailureCnt(1);
+            this.metricPublisher.publish(measurement); //signal that there was a failure
+            throw new JMXConnectionException("Exception during creating JMX connection for operation: " + taskType.name(), e);
+        }
+
+        try {
+            result = manageTask(connMgr);
```
vinaykumarchella: More applicable name suggestion - `manageTask` --> `runTaks`

priam/src/main/java/com/netflix/priam/cluster/management/IClusterManagement.java
```diff
@@ -15,12 +15,72 @@
...
+            measurement.incrementSuccessCnt(1);
+            this.metricPublisher.publish(measurement); //signal that there was a success
+            logger.info("Successfully finished executing the cluster management task: {} with result: {}", taskType, result);
+        } catch (Exception e) {
+            measurement.incrementFailureCnt(1);
+            this.metricPublisher.publish(measurement); //signal that there was a failure
+            throw new RuntimeException(String.format("Exception during %s.",taskType), e);
+
+        } finally {
+            if (connMgr != null)
```
vinaykumarchella: To simplify the resource management, `try-with-resources` would be a better alternative. 
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

priam/src/main/java/com/netflix/priam/cluster/management/SchemaConstant.java
```diff
@@ -0,0 +1,25 @@
+package com.netflix.priam.cluster.management;
+
+/**
+ * Created by aagrawal on 3/6/18.
+ */
+public class SchemaConstant {
+    public static final String SYSTEM_KEYSPACE_NAME = "system";
```
vinaykumarchella: If these constants are used only once, do you need these constants? also, instead of having these many variables, can you try any regex? or more common utility method `isSystemKeysapce` so that you can use it across the board in Priam if you need it in future?

priam/src/main/java/com/netflix/priam/cluster/management/TaskException.java
```diff
@@ -0,0 +1,15 @@
+package com.netflix.priam.cluster.management;
+
+/**
+ * Created by aagrawal on 2/28/18.
+ */
+public class TaskException extends Exception{
```
vinaykumarchella: What extra functionality is this giving apart from regular Exception? Are you looking for any specific TaskExceptions to act differently? I dont see that in code.

priam/src/main/java/com/netflix/priam/resources/CassandraAdmin.java
```diff
@@ -138,75 +141,45 @@ public Response cassRing(@PathParam("id") String keyspace) throws IOException, I
...
+    @GET
+    @Path("/compact")
+    public Response cassCompact() throws IOException, ExecutionException, InterruptedException {
+        JSONObject rootObj = new JSONObject();
 
-        List<String> flushed = null;
         try {
-            IClusterManagement task = new Flush(this.config, connMgr);
-            flushed = task.execute();
+            compaction.execute();
```
vinaykumarchella: Do you care or take care of scheduled compaction is running and calling this REST endpoint causes to start compaction while there is one already running.

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
```diff
@@ -271,7 +271,10 @@ public void restore(Date startTime, Date endTime) throws Exception {
             instanceState.setRestoreStatus(Status.FINISHED);
 
             //Start cassandra if restore is successful.
-            cassProcess.start(true);
+            if (!config.doesCassandraStartManually())
```
vinaykumarchella: Why is this change in this PR? Is it related or clearing backlog

priam/src/test/groovy/com/netflix/priam/cluser/management/TestCompaction.groovy
```diff
@@ -0,0 +1,81 @@
+package com.netflix.priam.cluser.management
```
vinaykumarchella: Can you include tests to verify the concurrent executions?

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
arunagrawal84: This was one of the misses we had w.r.t. restore. There is already a Jira for this by @jolynch

priam/src/main/java/com/netflix/priam/resources/CassandraAdmin.java
arunagrawal84: This is a REST endpoint where it is doing something requested by "admin". I don't think we should check (or change existing behavior) for any existing running compaction.

vinaykumarchella: How would you avoid admin issuing REST command while there is already a scheduled compaction running? One suggestion is to either queue the task or reject the REST endpoint request stating that there is already a running compaction and admin can retry at a later point in time.

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
```diff
@@ -0,0 +1,169 @@
...
+    public Compaction(IConfiguration config, IMetricPublisher metricPublisher, CassandraOperations cassandraOperations) {
+        super(config, Task.COMPACTION, metricPublisher, new CompactionMeasurement());
+        this.config = config;
+        this.cassandraOperations = cassandraOperations;
+    }
+
+    public final Map<String, List<String>> updateCompactionCFList(IConfiguration config) throws Exception {
+        final Map<String, List<String>> columnFamilyFilter = new HashMap<>(); //key: keyspace, value: a list of CFs within the keyspace
+
+        if (config.getCompactionCFList() == null || config.getCompactionCFList().isEmpty()) {
```
vinaykumarchella: When no filters are provided, can we just call `nt compact` equivalent, so that C* would take care of running parallel compactions as needed and as the #concurrent threads available for it to compact?

priam/src/main/java/com/netflix/priam/IConfiguration.java
```diff
@@ -213,6 +213,27 @@
...
+        return "-1";
+    }
+
+   /**
+    * Column Family(ies), comma delimited, to start compactions.
+    * Note:  the expected format is keyspace.cfname. If no value is provided then compaction is scheduled for all KS,CF(s)
+    *
+    * @return Column Family(ies), comma delimited, to start compactions.  If no filter is applied, returns null.
+    */
+    default String getCompactionCFList(){
```
vinaykumarchella: Do you want to consider exclude list and include list, so that on multi-tenant cluster, when there are 100's of CFs exists, it would be difficult to manage this list to exclude just few CFs.

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
```diff
@@ -0,0 +1,169 @@
...
+            if (cassandraOperations != null)
+                cassandraOperations.getKeyspaces().forEach(keyspaceName -> {
+                List<String> existingCfs = columnFamilyFilter.getOrDefault(keyspaceName, new ArrayList<>());
+                columnFamilyFilter.put(keyspaceName, existingCfs);
+            });
+            return columnFamilyFilter;
+        }
+
+        logger.info("Compaction: Override provided by user. Calculating KS/CF's to qualify for compaction");
+        String[] filters = config.getCompactionCFList().split(",");
```
vinaykumarchella: When CFList is such that all CFs in keyspace are included, it would be better if you let C* decide the parallelism based on its concurrent compactors, so that compactions would go faster, to achive that, you can just issue keyspace level compaction and let C* do the job.

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
```diff
@@ -0,0 +1,169 @@
...
+                logger.error("Keyspace: {} configured for compaction does not exist!", keyspace);
+                continue;
+            }
+
+            if (SchemaConstant.isSystemKeyspace(keyspace)) //no need to compact system keyspaces.
+                continue;
+
+            try {
+                if (columnfamilies == null || columnfamilies.isEmpty())
+                    cassandraOperations.forceKeyspaceCompaction(keyspace,  null);
```
vinaykumarchella: When would you get into this code path, considering the updateCompactionCFList logic above?

priam/src/main/java/com/netflix/priam/cluster/management/Flush.java
```diff
@@ -16,75 +16,83 @@
 package com.netflix.priam.cluster.management;
```
vinaykumarchella: It would be easier to review if you separate noncompaction related changes with this PR.

priam/src/main/java/com/netflix/priam/merics/CompactionMeasurement.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+package com.netflix.priam.merics;
+
+/**
+ * Measurement class for scheduled compactions
+ * Created by aagrawal on 2/28/18.
+ */
+
+public class CompactionMeasurement implements IMeasurement<Object> {
```
vinaykumarchella: We should start thinking about using spectator instead of our own measurements.

priam/src/main/java/com/netflix/priam/cluster/management/Flush.java
arunagrawal84: This was per design. Flush and compaction both extend IClusterManagement. This was done so we don't have repeat of code.

priam/src/main/java/com/netflix/priam/cluster/management/IClusterManagement.java
arunagrawal84: Changed this to String as we really never use the result as such. Too much code wrapping for nothing is not a good idea. If we ever decide to use the output in any other way (other than printing) we will do this.

priam/src/main/java/com/netflix/priam/cluster/management/TaskException.java
arunagrawal84: +1 Deleting. Since we already handle the counter increment in IClusterManagement. This does not make sense.

priam/src/main/java/com/netflix/priam/merics/CompactionMeasurement.java
arunagrawal84: I will open a separate JIRA for that as it will touch more components.

priam/src/main/java/com/netflix/priam/resources/CassandraAdmin.java
arunagrawal84: Resolved in IClusterManagement.

priam/src/test/groovy/com/netflix/priam/cluser/management/TestCompaction.groovy
arunagrawal84: Done.

priam/src/main/java/com/netflix/priam/cluster/management/Flush.java
vinaykumarchella: Correct, but the refactoring of IClusterManager could have been in a different commit.

priam/src/main/java/com/netflix/priam/merics/CompactionMeasurement.java
vinaykumarchella: 👍  We can go iterative on that, as we are touching any classes we can fix it.

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
arunagrawal84: If there are no filters provided, we just empty arraylist for the CF's. Later, we call compact on each keyspace (so we can avoid system keyspace). 
```List<String> existingCfs = columnFamilyFilter.getOrDefault(keyspaceName, new ArrayList<>());```

File: priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java [hunk 15]
```diff
@@ -0,0 +1,169 @@
...
+                String keyspaceName = filter[0];
+                String columnFamilyName = filter[1];
+
+                if (columnFamilyName.indexOf("-") != -1)
+                    columnFamilyName = columnFamilyName.substring(0, columnFamilyName.indexOf("-"));
+
+                logger.info("Compaction: Adding CF: {}.{}", keyspaceName, columnFamilyName);
+
+                List<String> existingCfs = columnFamilyFilter.getOrDefault(keyspaceName, new ArrayList<>());
+                if (!columnFamilyName.equalsIgnoreCase("*"))
```
arunagrawal84: In reference to your comment above, if all the CF's of a keyspace are included, we call compaction on that keyspace.

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
arunagrawal84: Actually it is already there, if you see line 87 (I added comment there)

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
vinaykumarchella: No, dont pass System keyspaces for compaction. 👍

priam/src/main/java/com/netflix/priam/cluster/management/Compaction.java
vinaykumarchella: Did you test passing `null` would not break anything?

vinaykumarchella: Did you test when passing `null` to forceKeyspaceCompaction?

arunagrawal84: Yes. CassandraOperations usually take care of 'null' inputs.

PR #677, opened 2018-06-06
change the default location of backup verification file and backup.status

arunagrawal84:


priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
```diff
@@ -1087,7 +1088,7 @@ public String getFlushInterval() {
 
     @Override
     public String getBackupStatusFileLoc() {
-        return config.get(CONFIG_BACKUP_STATUS_FILE_LOCATION,  "backup.status");
+        return config.get(CONFIG_BACKUP_STATUS_FILE_LOCATION,  getDataFileLocation() + File.separator + "backup.status");
```
vinaykumarchella: Why do you need `backup.status` here?

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -114,11 +115,11 @@ public BackupVerificationResult verifyBackup(List<BackupMetadata> metadata, Date
         //Download meta.json from backup location and uncompress it.
         List<String> metaFileList = new ArrayList<>();
         try {
-            bkpStatusFs.download(metas.get(0), new FileOutputStream("tmp_meta.json"));
+            bkpStatusFs.download(metas.get(0), new FileOutputStream(config.getDataFileLocation() + File.separator + "tmp_meta.json"));
```
jolynch: Can you use https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html to make this cleaner?

e.g. https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html#get(java.net.URI)

priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
arunagrawal84: This is an OSS implementation where we save backup/snapshot status in a file.

arunagrawal84: Since we saw that default location of this file was also in current working directory, we decided to change it to data folder for sanity purposes.

vinaykumarchella: 👍

PR #681, opened 2018-06-22
PostRestoreHookChanges

sumanth-pasupuleti:
Adding ability to call a post restore hook once files get downloaded as part of the restore process.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
+++ b/priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
@@ -1,8 +1,5 @@
 package com.netflix.priam.restore;
 
-<<<<<<< Updated upstream
-public class PostRestoreHook {
-=======
 import com.netflix.priam.IConfiguration;
 import com.netflix.priam.scheduler.NamedThreadPoolExecutor;
 import com.netflix.priam.utils.RetryableCallable;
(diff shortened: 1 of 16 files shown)


priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
```diff
@@ -0,0 +1,158 @@
...
+ * An implementation of IPostRestoreHook. Kicks off a child process for post restore hook using ProcessBuilder; uses heart beat monitor to monitor progress of the sub process
+ * and uses a file lock to pass the active state to the sub process
+ */
+public class PostRestoreHook implements IPostRestoreHook {
+    private static final Logger logger = LoggerFactory.getLogger(PostRestoreHook.class);
+    private final IConfiguration config;
+    private final Sleeper sleeper;
+    private final int ThreadWaitTimeInMs = 60000;
+    private final int HeartBeatTimeOutInMs = 60000;
+    private final int SubProcessTimeOutInDays = 2;
```
arunagrawal84: Does it make sense to make this configurable?

arunagrawal84: Also, once done, can you update the wiki to reflect the changes.

sumanth-pasupuleti: I like the idea of making the subprocesstimeout configurable - comes in handy when restore data size varies

PR #682, opened 2018-06-22
PostRestoreHook changes

sumanth-pasupuleti:
Adding ability to call a post restore hook once files get downloaded as part of the restore process.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
@@ -191,8 +191,7 @@ public class PriamConfiguration implements IConfiguration {
     private static final String CONFIG_POST_RESTORE_HOOK = PRIAM_PRE + ".postrestorehook";
     private static final String CONFIG_POST_RESTORE_HOOK_HEARTBEAT_FILENAME = PRIAM_PRE + ".postrestorehook.heartbeat.filename";
     private static final String CONFIG_POST_RESTORE_HOOK_DONE_FILENAME = PRIAM_PRE + ".postrestorehook.done.filename";
-
-
+    private static final String CONFIG_POST_RESTORE_HOOK_TIMEOUT_IN_DAYS = PRIAM_PRE + ".postrestorehook.timeout.in.days";
 
     //Running instance meta data
     private String RAC;
@@ -1159,5 +1158,10 @@ public class PriamConfiguration implements IConfiguration {
     @Override
     public String getPostRestoreHookDoneFileName() {
         return config.get(CONFIG_POST_RESTORE_HOOK_DONE_FILENAME, getDataFileLocation() + File.separator + "postrestorehook_done");
+    }
+
+    @Override
+    public int getPostRestoreHookTimeOutInDays() {
+        return config.get(CONFIG_POST_RESTORE_HOOK_TIMEOUT_IN_DAYS, 2);
     }
 }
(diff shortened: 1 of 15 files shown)


priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
```diff
@@ -0,0 +1,157 @@
...
+                        if (!process.waitFor(config.getPostRestoreHookTimeOutInDays(), TimeUnit.DAYS)) {
+                            logger.info("PostRestoreHook process did not complete within {} days. Forcefully terminating the process.", config.getPostRestoreHookTimeOutInDays());
+                            process.destroyForcibly();
+                        }
+
+                        if (process.exitValue() == 0) {
+                            logger.info("PostRestoreHook process completed successfully");
+                            break;
+                        }
+                        logger.info("PostRestoreHook process exited unsuccessfully");
```
vinaykumarchella: @sumanth-pasupuleti  Should it warn or error?

priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
```diff
@@ -0,0 +1,157 @@
...
+                    }
+                    logger.info("Completed PostRestoreHook execution");
```
vinaykumarchella: This leads to too verbose logging, consider changing this to debug

priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
```diff
@@ -187,6 +186,12 @@
     private static final String CONFIG_VPC_ROLE_ASSUMPTION_ARN = PRIAM_PRE + ".vpc.roleassumption.arn";
     private static final String CONFIG_DUAL_ACCOUNT = PRIAM_PRE + ".roleassumption.dualaccount";
 
+    //Post Restore Hook
+    private static final String CONFIG_POST_RESTORE_HOOK_ENABLED = PRIAM_PRE + ".postrestorehook.enabled";
+    private static final String CONFIG_POST_RESTORE_HOOK = PRIAM_PRE + ".postrestorehook";
+    private static final String CONFIG_POST_RESTORE_HOOK_HEARTBEAT_FILENAME = PRIAM_PRE + ".postrestorehook.heartbeat.filename";
+    private static final String CONFIG_POST_RESTORE_HOOK_DONE_FILENAME = PRIAM_PRE + ".postrestorehook.done.filename";
+    private static final String CONFIG_POST_RESTORE_HOOK_TIMEOUT_IN_DAYS = PRIAM_PRE + ".postrestorehook.timeout.in.days";
```
vinaykumarchella: Should it be days? or hours?

priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
sumanth-pasupuleti: good point. will change it to warn, since we are going to retry anyways

priam/src/main/java/com/netflix/priam/restore/PostRestoreHook.java
sumanth-pasupuleti: makes sense. will change it to debug

priam/src/main/java/com/netflix/priam/defaultimpl/PriamConfiguration.java
sumanth-pasupuleti: will keep it to days for now, but if we think it has to be a lower unit in the future based on the runs we'll have, will change it.

PR #691, opened 2018-07-16
Snapshot meta service

arunagrawal84:
This service will generate metafile containing the list of the files on the filesystem every hour. It will interact with Backup Upload/Download Service (BUDS) to ensure all the files are uploaded.


priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -201,40 +200,4 @@ public boolean isValidBackupDir(File keyspaceDir, File columnFamilyDir, File bac
      * Adds Remote path to the list of Remote Paths
      */
     protected abstract void addToRemotePath(String remotePath);
-
-    @Override
```
arunagrawal84: We were not using this as BackupEvent is done by lower layer (S3BaseFileSystem) for SNS notifications. Cleanup of dead code.

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -157,39 +154,41 @@ protected final void initiateBackup(String monitoringFolder, BackupRestoreUtil b
...
-                if (!isValidBackupDir(keyspaceDir, columnFamilyDir, backupDir)) {
+                if (!isValidBackupDir(keyspaceDir, backupDir)) {
                     continue;
                 }
 
-                String dirName = columnFamilyDir.getName();
-                String columnFamilyName = dirName.split("-")[0];
+                String columnFamilyName = columnFamilyDir.getName().split("-")[0];
 
-                if (backupRestoreUtil.isFiltered(BackupRestoreUtil.DIRECTORYTYPE.KEYSPACE, keyspaceDir.getName()) || //keyspace is filtered
```
arunagrawal84: Moved this as part of common library

priam/src/main/java/com/netflix/priam/compress/ICompression.java
```diff
@@ -25,6 +25,11 @@
 
 @ImplementedBy(SnappyCompression.class)
 public interface ICompression {
+
+    enum CompressionAlgorithm {
```
arunagrawal84: This is for future purposes only. Not used today.

priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
```diff
@@ -86,7 +86,7 @@ public void parseRemote(String remoteFilePath) {
         token = pieces.get(3);
         time = parseDate(pieces.get(4));
         type = BackupFileType.valueOf(pieces.get(5));
-        if (type != BackupFileType.META && type != BackupFileType.CL) {
+        if (type != BackupFileType.META && type != BackupFileType.META_V2 && type != BackupFileType.CL) {
```
sumanth-pasupuleti: nit: extract this into a method (maybe isDataFile()) since this is used in multiple places

arunagrawal84: Done

priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriter.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Delete the old meta files, if any present in the metaFileDirectory
+     */
+    public void cleanupOldMetaFiles(){
+        IOFileFilter fileNameFilter = FileFilterUtils.and(FileFilterUtils.prefixFileFilter(MetaFileInfo.META_FILE_PREFIX),
+                FileFilterUtils.or(FileFilterUtils.suffixFileFilter(MetaFileInfo.META_FILE_SUFFIX),
+                        FileFilterUtils.suffixFileFilter(MetaFileInfo.META_FILE_SUFFIX + ".tmp")));
+        Collection<File> files = FileUtils.listFiles(metaFileDirectory.toFile(), fileNameFilter, null);
+        files.stream().filter(file -> file.isFile()).forEach(file -> file.delete());
```
vinaykumarchella: Might want to add a `warn/debug` here mentioning deletion of metafiles, will be helpful in debugging.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,234 @@
...
+            logger.info("Initializaing SnapshotMetaService");
+            snapshotName = generateSnapshotName();
+
+            //Perform a cleanup of old snapshot meta_v2.json files, if any, as we don't want our disk to be filled by them.
+            //These files may be leftover
+            // 1) when Priam shutdown in middle of this service and may not be full JSON
+            // 2) No permission to upload to backup file system.
+            metaFileWriter.cleanupOldMetaFiles();
+
+            //TODO: enque all the old backup folder for upload/delete, if any, as we don't our disk to be filled by them.
```
vinaykumarchella: `enqueue` and `if any, as we don't **want** our disk to be filled by them.`

priam/src/main/java/com/netflix/priam/defaultimpl/CassandraOperations.java
```diff
@@ -0,0 +1,57 @@
...
+    private IConfiguration configuration;
+
+    @Inject
+    CassandraOperations(IConfiguration configuration)
+    {
+        this.configuration = configuration;
+    }
+
+    public void takeSnapshot(final String snapshotName) throws Exception {
+        new RetryableCallable<Void>() {
```
vinaykumarchella: Do you want this to have 15 retries? Also think about multiple threads calling this method since this is called from 2 places already.

priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriter.java
```diff
@@ -0,0 +1,155 @@
...
+ * This class will help in generation of meta.json files. This will encapsulate all the SSTables that were there on the file system.
+ * Created by aagrawal on 6/12/18.
+ */
+public class MetaFileWriter {
+    private static final Logger logger = LoggerFactory.getLogger(MetaFileWriter.class);
+    private final Provider<AbstractBackupPath> pathFactory;
+    private final IBackupFileSystem backupFileSystem;
+
+    private MetaFileInfo metaFileInfo;
+    private JsonWriter jsonWriter;
```
vinaykumarchella: You might want to add a comment here on the usage of JsonWriter (Streaming/ not intend to load the JSON in memory), otherwise, it would be tough to justify the usage considering the code complexity to build a JSON and ordering of method calls needed without enough context.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,234 @@
...
+
+        if (!StringUtils.isEmpty(cronExpression) && cronExpression.equalsIgnoreCase("-1")) {
+            logger.info("Skipping SnapshotMetaService as SnapshotMetaService cron is disabled via -1.");
+        } else {
+            if (StringUtils.isEmpty(cronExpression) || !CronExpression.isValidExpression(cronExpression))
+                throw new Exception("Invalid CRON expression: " + cronExpression +
+                        ". Please use -1, if you wish to disable SnapshotMetaService else fix the CRON expression and try again!");
+
+            cronTimer = new CronTimer(JOBNAME, cronExpression);
+            logger.info("Starting SnapshotMetaService with CRON expression {}", cronTimer.getCronExpression());
```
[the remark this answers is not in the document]
arunagrawal84: Well if we are there, we are definitely starting the service. If it is disabled, we never go in else loop and if it is invalid arg, we throw an exception.

vinaykumarchella:
fixNship from our offline conversation.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,227 @@
...
+                String prefix = PrefixGenerator.getSSTFileBase(file.getName());
+
+                if (prefix == null && file.getName().equalsIgnoreCase(CASSANDRA_MANIFEST_FILE))
+                    prefix = "manifest";
+
+                if (prefix == null)
+                    logger.error("Unknown file type with no SSTFileBase found: ", file.getAbsolutePath());
+
+                FileUploadResult fileUploadResult = FileUploadResult.getFileUploadResult(keyspace, columnFamily, file);
+                filePrefixToFileMap.putIfAbsent(prefix, new ArrayList<>());
```
hashbrowncipher: Do we interact with the HashMap on each iteration because we expect to see the same SSTable prefix in multiple directories? If not, I think this code could be simpler if we built the entire list and then add it to the Map in one call to `put()`.

priam/src/main/java/com/netflix/priam/backupv2/PrefixGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+    PrefixGenerator(IConfiguration configuration, InstanceIdentity instanceIdentity) {
+        this.configuration = configuration;
+        this.instanceIdentity = instanceIdentity;
+    }
+
+    public Path getPrefix() {
+        return Paths.get(configuration.getBackupLocation(), configuration.getBackupPrefix(), getAppNameReverse(), instanceIdentity.getInstance().getToken());
+    }
+
+    public Path getSSTPrefix() {
```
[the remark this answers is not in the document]
arunagrawal84: Yes, this will be used in stage 3 to actually determine where files would be uploaded.

priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriterBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        /**
+         * Add {@link ColumnfamilyResult} after it has been processed so it can be streamed to meta.json. Streaming write to meta.json is required so we don't get Priam OOM.
+         *
+         * @param columnfamilyResult a POJO encapsulating the column family result
+         * @throws IOException
+         */
+        public MetaFileWriterBuilder.DataStep addColumnfamilyResult(ColumnfamilyResult columnfamilyResult) throws IOException {
+            if (jsonWriter != null && columnfamilyResult != null)
```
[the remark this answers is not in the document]
arunagrawal84: Reasonable, even if there are no SSTables for a CF, we would have an empty tag for that CF.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,227 @@
...
+        logger.debug("Scanning for all SSTables in: {}", snapshotDir.getAbsolutePath());
+
+        Map<String, List<FileUploadResult>> filePrefixToFileMap = new HashMap<>();
+        Collection<File> files = FileUtils.listFiles(snapshotDir, FileFilterUtils.fileFileFilter(), null);
+
+        files.stream().filter(file -> file.exists()).filter(file -> file.isFile()).forEach(file -> {
+            try {
+                String prefix = PrefixGenerator.getSSTFileBase(file.getName());
+
+                if (prefix == null && file.getName().equalsIgnoreCase(CASSANDRA_MANIFEST_FILE))
```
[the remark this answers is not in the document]
arunagrawal84: The goal of SnapshotMetaService is to capture filesystem at a moment in time. It will though "try" to push those files for upload via BackupUploadDownloadService (in next version). It would be completely ok to fail to enqueue the files to upload as there will be another service pushing all the files in the queue for upload (think of this like a cleanup thread and will help us in "resuming" any failed backup for any reason).

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,227 @@
...
+                    prefix = "manifest";
+
+                if (prefix == null)
```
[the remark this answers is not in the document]
arunagrawal84: I have made to exclude such files from upload. If we really don't understand the file we should not be uploading. An example would be user dropping some file/folder structure in data folder mistakenly.  The only thing we need to ensure would be this method works for `future` version of C* file name changes. Right now logic expects a certain nomenclature of file names and works for 2.x and 3.x C* version.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -0,0 +1,227 @@
...
+            columnfamilyResult.addSstable(ssTableResult);
+        });
+        return columnfamilyResult;
+    }
+
+    @Override
+    protected void processColumnFamily(final String keyspace, final String columnFamily, final File backupDir) throws Exception {
+        File snapshotDir = getValidSnapshot(backupDir, snapshotName);
+        // Process this snapshot folder for the given columnFamily
+        if (snapshotDir == null) {
```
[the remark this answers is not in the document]
arunagrawal84: Short answer: no, this method will ensure that all active CF are accounted for. 
Long version: 
Before we execute this command, we have asked C* to take a snapshot which means a valid snapshot for all the "active" CF should exist. 
We walk over folder structure in C* data folder and there could be "folders" in data folder which are not really CF or some leftover from previous, user errors etc.  So we really need this to avoid such bad case. 
Also, this method is per CF data folder so if it really is not an active CF, we skipping that CF is valid.

PR #693, opened 2018-07-18
More resilient incrementals

jolynch:
This change makes it so Priam will continue re-trying incremental uploads forever (as opposed to only retrying them the next time that you restart Priam).

Before we could fail to upload incrementals with 1.5s of AWS S3 unavailability. Now we can only fail incrementals if we're misconfigured or AWS is actually permanently down. The existing controls for queue size are sufficient to control backpressure here.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/parallel/ITaskQueueMgr.java
+++ b/priam/src/main/java/com/netflix/priam/backup/parallel/ITaskQueueMgr.java
@@ -26,19 +26,26 @@ package com.netflix.priam.backup.parallel;
  */
 public interface ITaskQueueMgr<E> {
 
+    /**
+     * Adds the provided task into the queue if it does not already exist. For performance reasons
+     * this is best effort and therefore callers are responsible for handling duplicate tasks.
+     *
+     * This method will block if the queue of tasks is full
+     * @param task The task to put onto the queue
+     */
     void add(E task);
 
-    /*
+    /**
      * @return task, null if none is available.
      */
     E take() throws InterruptedException;
 
-    /*
+    /**
      * @return true if there are tasks within queue to be processed; false otherwise.
      */
     Boolean hasTasks();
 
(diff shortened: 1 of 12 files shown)


priam/src/main/java/com/netflix/priam/backup/parallel/ITaskQueueMgr.java
```diff
@@ -26,31 +26,53 @@
...
+     * This method will block if the queue of tasks is full
+     * @param task The task to put onto the queue
+     */
+    void add(E task);
+
+    /**
+     * Adds the provided task into the queue if it does not already exist. For performance reasons
+     * this is best effort and therefore callers are responsible for handling duplicate tasks.
+     *
+     * This method should not block. If no implementation is provided however this is equivalent to
```
arunagrawal84: nit: The default behavior will be to block as default impl is to block using "add" and then return true. I understand the behavior is modified in CassandraBackupQueueMgr though.

jolynch: Yea, I've clarified more.

priam/src/main/java/com/netflix/priam/backup/parallel/IncrementalConsumer.java
```diff
@@ -76,6 +79,9 @@ public Void retriableCall() throws Exception {
                         if (is == null) {
                             throw new NullPointerException("Unable to get handle on file: " + bp.getFileName());
                         }
+                        // Important context: this upload call typically has internal retries but those are only
```
arunagrawal84, line 82: Filesystem upload does not have retries AFAIK for files which are large. Only small files have retries. The only retry coming here is from RetryCallable used just above for those big files.

arunagrawal84:
Also, we would need PR for 3.11 branch.

PR #697, opened 2018-07-31
Fix G1GC Tuning

jolynch:
G1GC should never be run with a young gen heap size, so this change sets
the young gen to comment out the young gen heap all together for G1GC
enabled clusters


arunagrawal84:
LGTM

PR #699, opened 2018-08-02
Compactions on CRON

arunagrawal84:


priam/src/main/java/com/netflix/priam/defaultimpl/CassandraOperations.java
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        }.call();
+    }
+
+    public void forceKeyspaceCompaction(String keyspaceName, String... columnfamilies) throws Exception{
+        new RetryableCallable<Void>(){
+            public Void retriableCall() throws Exception{
+                try(JMXNodeTool nodeTool = JMXNodeTool.instance(configuration)) {
+                        nodeTool.forceKeyspaceCompaction(keyspaceName, columnfamilies);
```
arunagrawal84: JMX interface has changed between cassandra 2.1.17 (pinned version) and c* 3.x. We need to get the 3.x cassandra-all for compactions to actually proceed.

PR #701, opened 2018-08-03
3.11 postrestorehook timeout config

sumanth-pasupuleti:


priam/src/main/java/com/netflix/priam/IConfiguration.java
```diff
@@ -779,4 +779,16 @@
      * @return time out for post restore hook in days
      */
     public int getPostRestoreHookTimeOutInDays();
+
+    /**
+     * Heartbeat timeout (in ms) for post restore hook
+     * @return heartbeat timeout for post restore hook
+     */
+    public int getPostRestoreHookHeartBeatTimeoutInMs();
```
arunagrawal84: please provide a default value in the interface, so we don't need to copy this method in all implementations of IConfiguration. Also, drop the `public` modifier as the interface is already `public`.

priam/src/test/java/com/netflix/priam/FakeConfiguration.java
```diff
@@ -914,4 +914,12 @@ public String getPostRestoreHookDoneFileName() {
     public int getPostRestoreHookTimeOutInDays() {
         return 2;
     }
+
+    @Override
+    public int getPostRestoreHookHeartBeatTimeoutInMs() {
```
arunagrawal84: No need to provide this if you provide default implementation in IConfiguration.

sumanth-pasupuleti:
Thanks @arunagrawal84 . will be incorporating your comments and sending out a new PR (my branch ended up in a messy state)

sumanth-pasupuleti:
Closing this PR and sending out another from a clean branch.

PR #706, opened 2018-08-14
Move to spectator to collect metrics

arunagrawal84:


sumanth-pasupuleti:
From the outset, it looks good to me. I assume we are migrating 'all' Priam metrics to spectator

PR #712, opened 2018-08-28
Forgotten files

arunagrawal84:
Forgotten files: Increment metric when a forgotten file is detected in C* data folder.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
+++ b/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
@@ -33,13 +33,15 @@ import com.netflix.priam.utils.DateUtil;
 import com.netflix.priam.utils.ThreadSleeper;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.filefilter.FileFilterUtils;
+import org.apache.commons.io.filefilter.IOFileFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.*;
-import java.util.stream.Collectors;
+import java.util.regex.Pattern;
 
 /**
  * Task for running daily snapshots
@@ -61,6 +63,8 @@ public class SnapshotBackup extends AbstractBackup {
     private List<AbstractBackupPath> abstractBackupPaths = null;
     private CassandraOperations cassandraOperations;
     private BackupMetrics backupMetrics;
+    private final String TMP_EXT = ".tmp";
+    private final Pattern tmpFilePattern = Pattern.compile("^((.*)\\-(.*)\\-)?tmp(link)?\\-((?:l|k).)\\-(\\d)*\\-(.*)$");
 
     @Inject
     public SnapshotBackup(IConfiguration config, Provider<AbstractBackupPath> pathFactory,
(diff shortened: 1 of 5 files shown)


priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -180,11 +190,50 @@ private void notifyObservers() {
...
+            }).collect(Collectors.toList());
+
+            forgottenFiles.parallelStream().forEach(file -> logger.info("Forgotten file: {} for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName()));
+
+            if (forgottenFiles == null || forgottenFiles.size() == 0)
+                return;
+
+            backupMetrics.incrementForgottenFiles(forgottenFiles.size());
+            logger.warn("Forgotten files: {} found for CF: {}", forgottenFiles.size(), columnfamilyDir.getName());
+        }catch (Exception e)
```
jolynch: nit: formatting

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -180,11 +190,50 @@ private void notifyObservers() {
...
+        {
+            //Eat the exception, if there, for any reason. This should not stop the snapshot for any reason.
```
jolynch: nit: formatting

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -180,11 +190,50 @@ private void notifyObservers() {
...
+            logger.error("Exception occurred while trying to find forgottenFile. Ignoring the error and continue with backup.", e);
```
jolynch: nit: readability: "Ignoring the error and continuing with remaining backup."

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -180,11 +190,50 @@ private void notifyObservers() {
...
+            for (File file : snapshotFiles) {
+                //Get its parent directory file based on this file.
+                File originalFile = new File(columnfamilyDir, file.getName());
+                columnfamilyFiles.remove(originalFile);
+            }
+
+            if (columnfamilyFiles.size() == 0)
+                return;
+
+            //There might be new SSTables since we took snapshot. Remove them from this list.
```
jolynch: I think you need to document more of the algorithm here. From what I can tell you're considering any file which is older than the snapshot instant to be forgotten. I don't think this is what you want. You probably want files which are older than gc_grace, not in the snapshot, and not -tmp- files.

I think the algorithm needs to

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -180,11 +190,50 @@ private void notifyObservers() {
...
-        if (null != snapshotDir)
-            abstractBackupPaths.addAll(upload(snapshotDir, BackupFileType.SNAP));
-        else
+
+        if (snapshotDir == null) {
             logger.warn("{} folder does not contain {} snapshots", backupDir, snapshotName);
+            return;
+        }
+
+        findForgottenFiles(snapshotDir);
```
jolynch: Is a metric the best way to deal with this, is the eventual plan to automate the movement of these files from the data directory to a `lost+found` directory or some such?

arunagrawal84: The eventual plan is to move (configurable). First step was to ensure we have metrics in place before Priam touch any SSTable.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
arunagrawal84: Why not? If a file which was written before the snapshot time is not part of the snapshot, isn't that forgotten (or tmp file)? I don't understand the need for GC_GRACE from your comments. 
Anything which is -tmp- 👍 : Yes I forgot about it as there could have a compaction in progress while snapshot was executed. We should remove any -tmp- file from CF for comparison purposes. Also, in reality, update time should be updated for -tmp- files if they are actively being written to but race condition while code execution (after snapshot) and compaction flush make me believe this should be done.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
jolynch: Cool, might want to leave that context in a comment somewhere or a github issue or something? For example I think Priam should require a force option to stop if it detects forgotten files ...

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
jolynch: Cassandra doesn't instantly delete sstables, especially as compaction or repair byproducts there can be sstables that are not in the view but maybe a few hours old. This is fine as long as the files are not older than gc_grace (in which case they could contain deleted data).

PR #713, opened 2018-09-01
First take at supporting Cassandra 4.x

jolynch:
So lots of things have broken in 4.x, most noticeably:
* EC2MRS needs to be tuned by Priam so we can set the legacy option (preventing renaming of racks/dcs)
* Thrift is gone, don't need to have all those options any more
* A few JMX endpoints went away
* jvm-options is a different file now
* InetAddressAndPort broke the seed provider

My plan is to cut a release for 4.0.0-alpha with this branch.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
+++ b/priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
@@ -28,6 +28,8 @@ import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import java.io.*;
+import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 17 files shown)


priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
```diff
@@ -201,12 +214,10 @@ protected void configureCommitLogBackups() throws IOException {
...
 
-        FileOutputStream fos = null;
-        try {
-            fos = new FileOutputStream(new File(config.getCommitLogBackupPropsFile()));
+        try (FileOutputStream fos = new FileOutputStream(new File(config.getCommitLogBackupPropsFile()))) {
             props.store(fos, "cassandra commit log archive props, as written by priam");
-        } finally {
-            IOUtils.closeQuietly(fos);
+        } catch (IOException e) {
+            logger.error("Could not store commitlog_archiving.properties", e);
```
arunagrawal84: Why are we eating exception here? Earlier priam would have been in restart loop but now it will just proceed with next steps?

jolynch: Ok, adding back the throw.

jolynch:
ok, merging this so we can start iterating on 4.x. We can re-visit if needed later.

PR #716, opened 2018-09-11
Provide an overwrite method to force Priam to replace a particular ip

jolynch:
This allows us to work around the A->B->C replacement problem where
Priam get's confused about who to replace


arunagrawal84:
LGTM

PR #722, opened 2018-09-18
Backup version 2.0 Milestone2 

arunagrawal84:
- Add local database support 
- refactored filesystem to get them ready for milestone 2
TODO: Integration of local database generation with filesystem `upload` is still remaining.


priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    public void writeLocalDB(final Path localDBFile, final LocalDB localDB) throws Exception {
+        if (localDB == null || localDBFile == null || localDBFile.toFile().isDirectory())
+            throw new Exception("Invalid Arguments: localDbFile: " + localDBFile + ", localDB: " + localDB);
+
+        if (!localDBFile.getParent().toFile().exists())
+            localDBFile.getParent().toFile().mkdirs();
+
+        try (FileWriter writer = new FileWriter(localDBFile.toFile())) {
```
hashbrowncipher: I think it would be worthwhile to make this update atomic, especially since this file will be updated so frequently.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+            writer.write(localDB.toString());
+        }
+    }
+
+    public LocalDB readLocalDB(final Path localDBFile) throws Exception {
+        //Verify it is file and it exists.
+        if (localDBFile.toFile().isDirectory())
```
hashbrowncipher: Do we need to throw this exception ourselves, or can we let Java throw an exception for us when we attempt to construct a FileReader below?

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+        return getLocalDBEntry(fileUploadResult, localDB);
+    }
+
+    private LocalDBEntry getLocalDBEntry(final FileUploadResult fileUploadResult, final LocalDB localDB) throws Exception {
+        if (localDB == null || localDB.getLocalDb().isEmpty())
+            return null;
+
+        //Get local db entry for same file and version.
+        List<LocalDBEntry> localDBEntries = localDB.getLocalDb().stream().filter(localDBEntry ->
+                (localDBEntry.getFileUploadResult().getFileName().toFile().getName().toLowerCase().equals(fileUploadResult.getFileName().toFile().getName().toLowerCase())))
```
hashbrowncipher: I'm having trouble reading this. I think it would be easier if all three filters were unified into a single filter, with a multiline anonymous function.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+                .filter(localDBEntry -> (localDBEntry.getFileUploadResult().getLastModifiedTime().equals(fileUploadResult.getLastModifiedTime())))
+                .filter(localDBEntry -> (localDBEntry.getFileUploadResult().getCompression().equals(fileUploadResult.getCompression())))
+                .collect(Collectors.toList());
+
+        if (localDBEntries == null || localDBEntries.isEmpty())
```
hashbrowncipher: I can't think of a way that `localDBEntries` would be null here.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+            });
+
+            if (logger.isDebugEnabled())
+                logger.debug("Local DB: {}", localDB);
+            return localDB;
+        }
+    }
+
+    static class LocalDB {
+        private List<LocalDBEntry> localDb;
```
hashbrowncipher: There are two different functions called getLocalDb with very different behavior. I think life for the reader would be better if this variable were called `entries` or somesuch.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+        public LocalDB(List<LocalDBEntry> localDb) {
+            this.localDb = localDb;
+        }
+
+        public List<LocalDBEntry> getLocalDb() {
+
+            return localDb;
+        }
+
+        public void setLocalDb(List<LocalDBEntry> localDb) {
```
hashbrowncipher: Is this ever called? Should we make the List final, and remove this method?

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return GsonJsonSerializer.getGson().toJson(this);
+        }
+    }
+
+    static class LocalDBEntry {
+        private FileUploadResult fileUploadResult;
```
hashbrowncipher: should this be final?

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+            writeLocalDB(localDBFile, localDB);
+        }
+
+        return localDB;
+    }
+
+    private LocalDB getLocalDB(final FileUploadResult fileUploadResult) throws Exception {
+        final Path localDbPath = getLocalDBPath(fileUploadResult);
+        //Retry for reading.
+        return new RetryableCallable<LocalDB>(5, 1000) {
```
hashbrowncipher: In what ways would readLocalDB fail that would require a retry? I'm looking, and I'm having trouble seeing anything other than local filesystem operations in this code path.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+            public LocalDB retriableCall() throws Exception {
+                return readLocalDB(localDbPath);
+            }
+        }.call();
+    }
+
+    public LocalDBEntry getLocalDBEntry(final FileUploadResult fileUploadResult) throws Exception {
+        LocalDB localDB = getLocalDB(fileUploadResult);
+
+        if (localDB == null || localDB.getLocalDb() == null || localDB.getLocalDb().isEmpty())
```
hashbrowncipher: It looks like we verify these attributes, and then call the `getLocalDBEntry(a, b)` method below. Could we do the verification once instead of twice?

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+                logger.debug("Local entry found: {}", localDBEntries.get(0));
+
+            return localDBEntries.get(0);
+        }
+
+        throw new Exception("Unexpected behavior: More than one entry found in local database for the same file. FileUploadResult: " + fileUploadResult);
+    }
+
+    public Path getLocalDBPath(final FileUploadResult fileUploadResult) {
+        return Paths.get(configuration.getDataFileLocation(), LOCAL_DB, fileUploadResult.getKeyspaceName(), fileUploadResult.getColumnFamilyName(), PrefixGenerator.getSSTFileBase(fileUploadResult.getFileName().toFile().getName()) + ".localdb");
```
hashbrowncipher: Would it be possible to wrap this line in a way that makes it easier to read?

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+        if (localDBEntry.getTimeLastReferenced() == null)
+            throw new Exception("Time last referenced in localDB can never be null");
+
+        if (localDBEntry.getBackupTime() == null)
+            throw new Exception("Backup time for localDB can never be null");
+
+        final Path localDBFile = getLocalDBPath(localDBEntry.getFileUploadResult());
+
+        if (!localDBFile.getParent().toFile().exists())
+            localDBFile.getParent().toFile().mkdirs();
```
hashbrowncipher: if you wanted to be [EAFP rather than LBYL](https://stackoverflow.com/a/405220), you could remove the .exists() check here and just call mkdirs unconditionally.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * This class is used to create a local DB entry for SSTable components. This will be used to identify when a version
+ * of a SSTable component was last uploaded or last referenced.
+ * This db entry will be used to enforce the TTL of a version of a SSTable component as we will not be relying on
+ * backup file system level TTL.
+ * Local DB should be copied over to new instance replacing this token in case of instance replacement. If no local DB
```
hashbrowncipher, line 42: Since we're going to need to have an operation that reconstitutes the localdb anyway, I think we should take a Crash Only Software approach and treat reconstitution as the primary operation, with the localdb functioning only as a cache. In that case, we could omit the on-disk component of the localdb entirely, and just keep the relevant information in memory.

All of the information in the localdb is easily gotten from S3, with the only exception being the timeLastReferenced. For that, we could take one of two approaches:
* diff manifests against each other, and determine timeLastReferenced from the first manifest where a given SSTable does not appear (although this is probably best renamed to firstNotReferenced)
* or do the above, and also add tombstone entries in S3 when we notice an sstable has disappeared. Deletion is then super easy: scan over objects and delete any whose tombstones are older than the TTL.

Since we have to do one of the above options anyway, I think getting rid of the on-disk component makes operations (node replacement) easier and reduces complexity within Priam.

hashbrowncipher (commented):
I left a bunch of inline review, but I think we should first answer the threshold question of why an on-disk localdb is useful, and whether we would instead do better with an in-memory cache of data that is stored in S3. (see comments on line 41)

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
arunagrawal84: 1. Every priam startup then would have to download all the meta files and scan over 100's of thousands of SSTables (via list operation). List operation is time-consuming and I prefer instant startups. 
2. I don't like creating a strong dependency on a remote file system being available on startup. 

We can though avoid above by doing in the background with the best effort with failback being upload everything (an expensive operation). 

In "replacement" nodes we can land in a similar situation where the remote database is not available but in this situation, we have to upload everything (as it is a new node - new SSTables). 

To ensure a proper cleanup we will always require a background thread where we reconstitute DB so we can catch up if the remote file system is available at later time. Thus, I don't get your point of reducing "complexity"

priam/src/main/java/com/netflix/priam/aws/S3EncryptedFileSystem.java
```diff
@@ -65,122 +61,68 @@ public S3EncryptedFileSystem(Provider<AbstractBackupPath> pathProvider, ICompres
...
-            RangeReadInputStream rris = new RangeReadInputStream(s3Client, getPrefix(config), path);
-
-        	/*
+    protected void downloadFileImpl(Path remotePath, Path localPath) throws BackupRestoreException {
+        try (OutputStream os = new FileOutputStream(localPath.toFile());
+             RangeReadInputStream rris = new RangeReadInputStream(s3Client, getPrefix(config), super.getFileSize(remotePath), remotePath.toString())
+        ) {
+            /*
              * To handle use cases where decompression should be done outside of the download.  For example, the file have been compressed and then encrypted.
         	 * Hence, decompressing it here would compromise the decryption.
```
jolynch, line 75: nit: replace tabs with spaces

priam/src/main/java/com/netflix/priam/aws/S3EncryptedFileSystem.java
```diff
@@ -65,122 +61,68 @@ public S3EncryptedFileSystem(Provider<AbstractBackupPath> pathProvider, ICompres
...
-        }
-
-        try {
+        File compressedDstFile = new File(localPath.toString() + ".compressed");
+        if (logger.isDebugEnabled())
+            logger.debug("Compressing {} with chunk size {}", compressedDstFile.getAbsolutePath(), chunkSize);
 
+        try (InputStream in = new FileInputStream(localPath.toFile());
+             BufferedOutputStream compressedBos = new BufferedOutputStream(new FileOutputStream(compressedDstFile))) {
             Iterator<byte[]> compressedChunks = this.compress.compress(in, chunkSize);
```
jolynch, line 100: I thought we were taking compression out of the picture? Are we just leaving it in to minimize scope of the change (I fully agree with that choice just wondering).

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -57,75 +59,48 @@ protected void setFileSystem(IBackupFileSystem fs) {
         this.fs = fs;
     }
 
+    private AbstractBackupPath getAbstractBackupPath(final File file, final BackupFileType type) throws Exception {
```
jolynch: I think this can only throw a ParseException, let's just throw that (no pokemon catching)

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -57,75 +59,48 @@ protected void setFileSystem(IBackupFileSystem fs) {
...
+
+
     /**
      * Upload files in the specified dir. Does not delete the file in case of
-     * error.  The files are uploaded serially.
+     * error.  The files are uploaded serially or async based on flag provided.
      *
      * @param parent Parent dir
      * @param type   Type of file (META, SST, SNAP etc)
+     * @param async  Upload the file(s) in async fashion if enabled.
```
jolynch, line 76: I don't think we ever want non async uploads? think the only tunable we need is the size of the backing consumer queue (i.e. if we only want a single file upload at a time we should make the backup thread pool size 1)

Put differently, let's just have async uploads always.

priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -99,13 +98,8 @@ else if (UpdateSecuritySettings.firstTimeUpdated)
 
             // Start the Incremental backup schedule if enabled
             if (config.isIncrBackup()) {
-                if (!config.isIncrBackupParallelEnabled()) {
                     scheduler.addTask(IncrementalBackup.JOBNAME, IncrementalBackup.class, IncrementalBackup.getTimer());
```
jolynch: nit: spacing

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -0,0 +1,205 @@
...
+    protected BackupMetrics backupMetrics;
+    private Set<Path> tasksQueued;
+    private ThreadPoolExecutor fileUploadExecutor;
+    private ThreadPoolExecutor fileDownloadExecutor;
+
+    @Inject
+    public AbstractFileSystem(IConfiguration configuration, BackupMetrics backupMetrics,
+                              BackupNotificationMgr backupNotificationMgr) {
+        this.backupMetrics = backupMetrics;
+        //Add notifications.
```
jolynch: general nit: space between // and the word

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -0,0 +1,205 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.FileNotFoundException;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.*;
+
+/**
+ * Created by aagrawal on 8/30/18.
```
jolynch, line 44: This seems to be the main logic of upload and download, can you fill in a bit of a class doc that explains what this class is responsible vs the implementations?

E.g. this is responsible for managing parallelism and orchestrating the upload and download, but the subclasses actually implement the details of uploading a file

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -0,0 +1,205 @@
...
+            return localPath;
+        });
+    }
+
+    @Override
+    public void uploadFile(final Path localPath, final Path remotePath, final AbstractBackupPath path, final int retry, final boolean deleteAfterSuccessfulUpload) throws FileNotFoundException, BackupRestoreException {
+        if (localPath == null || remotePath == null || !localPath.toFile().exists() || localPath.toFile().isDirectory())
+            throw new FileNotFoundException("File do not exist or is a directory. localPath: " + localPath + ", remotePath: " + remotePath);
+
+        if (!tasksQueued.contains(localPath)) {
```
jolynch: I'd just do `if (tasksQueued.add(localPath))`

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -0,0 +1,205 @@
...
+    private ThreadPoolExecutor fileUploadExecutor;
+    private ThreadPoolExecutor fileDownloadExecutor;
+
+    @Inject
+    public AbstractFileSystem(IConfiguration configuration, BackupMetrics backupMetrics,
+                              BackupNotificationMgr backupNotificationMgr) {
+        this.backupMetrics = backupMetrics;
+        //Add notifications.
+        this.addObserver(backupNotificationMgr);
+        tasksQueued = new HashSet<>(configuration.getBackupQueueSize());
```
jolynch: This probably needs to be a concurrent hash set. Also I think it's slightly odd to pre-size this (just it resize automatically)

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -24,111 +24,72 @@
...
+    public static final Map<String, List<String>> getFilter(String inputFilter) throws IllegalArgumentException {
+        if (StringUtils.isEmpty(inputFilter))
+            return null;
 
-        }
+        final Map<String, List<String>> columnFamilyFilter = new HashMap<>(); //key: keyspace, value: a list of CFs within the keyspace
 
-        if (configColumnfamilyFilter == null || configColumnfamilyFilter.isEmpty()) {
+        String[] filters = inputFilter.split(",");
+        for (int i = 0; i < filters.length; i++) { //process each filter
```
jolynch: nit: this comment doesn't add much imo

priam/src/main/java/com/netflix/priam/backup/IncrementalBackup.java
```diff
@@ -94,7 +94,7 @@ private void notifyObservers() {
 
     @Override
     protected void processColumnFamily(String keyspace, String columnFamily, File backupDir) throws Exception {
-        List<AbstractBackupPath> uploadedFiles = upload(backupDir, BackupFileType.SST);
+        List<AbstractBackupPath> uploadedFiles = upload(backupDir, BackupFileType.SST, config.enableAsyncIncremental());
```
jolynch, line 97: +1 to enabling async all the time.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
jolynch: I'm confused, don't we just need to download the single latest "localdb" file? We may re-do some work that has been already been done as we could get stale data but that seems fine?

Like I guess what I'm saying is whatever state this file is storing, we should just put it in S3 and then we can get the entire state (possibly slightly stale) from S3 in a single GET call. I think I agree with Josh that this way we actually practice recovery (not having this file) all the time, and ensure that the metadata we're storing in S3 is fast to reconstruct on machine failure.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
jolynch: Or split up onto multiple lines.

priam/src/main/java/com/netflix/priam/google/GoogleEncryptedFileSystem.java
```diff
@@ -274,59 +208,27 @@ public void upload(AbstractBackupPath path, InputStream in)
...
-    public long bytesUploaded() {
-        // TODO Auto-generated method stub
-        return 0;
-    }
-
-    @Override
-    public long getBytesUploaded() {
-        return 0;
+    protected long uploadFileImpl(Path localPath, Path remotePath) throws BackupRestoreException {
+        throw new UnsupportedOperationException();
```
jolynch, line 220: wait this is unsupported?

priam/src/test/java/com/netflix/priam/backup/FakeBackupFileSystem.java
```diff
@@ -156,13 +120,32 @@ public long getAWSSlowDownExceptionCounter() {
...
-
     }
 
-	@Override
-	public void download(AbstractBackupPath path, OutputStream os,
-			String diskPath) throws BackupRestoreException {
-      download(path, os);
-	}
+    @Override
+    protected void downloadFileImpl(Path remotePath, Path localPath) throws BackupRestoreException {
```
jolynch, line 126: nice!

jolynch (approved):
Fix and ship. I think my main two pieces of feedback are:

1. There should only be one option and that is async.
2. I think I agree with Josh that we should be designing this as crash only software, if we can't reconstruct the meta state quickly on startup then we're not storing the right metadata in s3 imo.

I don't view either of these as blockers though, given how large this change is I'm a fan of merging it and iterating (before the merge conflicts get too much worse)

priam/src/main/java/com/netflix/priam/aws/S3EncryptedFileSystem.java
arunagrawal84: Considering the current changes, I wanted that to come separately and only in 3.11 branch as Agesthius expects file to be compressed ATM.

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
arunagrawal84: I agree with that, however, I would feel way comfortable retiring sync uploads when we have more perf results with async only. 
Reason to add async to snapshot feature was to get all the numbers and in backup V2.0 async will be the only option.

priam/src/main/java/com/netflix/priam/google/GoogleEncryptedFileSystem.java
arunagrawal84: So far we have only supported download from Google File System (GFS). Hopefully, in future, we can support GFS in the same way as S3 (all operations) by hiding the complexity of encryption and compression.

priam/src/main/java/com/netflix/priam/backupv2/LocalDBReaderWriter.java
arunagrawal84: In current design "localdb" is created per SSTable i.e. it contains only its components. We do not upload this localdb at all to S3 (or we did not intend to as such) as it will change almost every time we do "full" snapshot (every hour in the new world) until it gets compacted out. Note that localdb contains the time a file was last referenced to implement the TTL. 

If I understand correctly, you propose to have a single "localdb" file on disk and upload that file to S3 at regular intervals. For recovery/restart, one should download that file and then do "list" and download of every meta to implement TTL. The worst that would happen during this period is priam uploading the same file again (which for the restart can be avoided by verifying remote file system)

Another approach could be pushing the "localdb" to a real database. We intentionally avoided dependency on another database for disaster recovery.

PR #724, opened 2018-09-25
Expose Priam config via HTTP

jolynch:
Expose Priam config via HTTP and persist to disk
    
    This provides two new HTTP APIs for external tools to access Priam
    configuration. These APIs allow external tools to work with Priam to
    manage Cassandra while still using Priam's flat configuration as a
    source of truth for the Cassandra instance
    
    GET /v1/config/structured/all -> returns all Priam IConfiguration fields
    
    GET /v1/config/unstructured/X?default=Y -> returns the property X, or Y if
    it is not found (404 will be raised if no default and the value doesn't
    exist)
    
    In addition to the HTTP apis there is also now a persister thread which
    makes it so that every minute (by default) configuration is written out
    to a directory in tmp by Priam. This way tools can access config even if
    Priam is dead. The file permissions are 600 so that only the Priam user
    can access it.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -16,11 +16,18 @@
  */
 package com.netflix.priam.config;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.ImplementedBy;
 import com.netflix.priam.identity.config.InstanceDataRetriever;
 import com.netflix.priam.scheduler.SchedulerType;
 import com.netflix.priam.scheduler.UnsupportedTypeException;
 
+import java.io.File;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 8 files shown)


hashbrowncipher (approved):
This looks great!

vinaykumarchella:
Suggestion: Different thought on this,  why not expose the IConfiguration as a JSON, if needed we can expose property level API from JSON, so that it supports `GET - /property/{propertyName:.+`. 

This has several benefits. 
1. We will be avoiding reflection in configuration class considering its performance costs
2. Avoid dealing with method name complexities
3. Avoid Config prefix confusion in oss and nf versions.

E.g., Exposing IConfiguration as JSON gives below output

```
REST/config/list

{
  "backupLocation": "foobar",
  "backupRetentionDays": 30,
  "gracefulDrainHealthWaitSeconds": 60,
  "backupPrefix" : "/backuplocation/foobar"
} 

REST/config/get/backupLocation

{
  "backupLocation": "foobar"
} 
```

jolynch:
Ok I took @vinaykumarchella 's advice and just made it `config/structured/<group>/<key?>`. So you can do `/REST/v1/config/structured/all` and `/REST/v1/config/structured/all/compactionThroughput` and it does the right thing.

There is also an unstructured api for direct property access and now we write out the structured config to `/tmp` as well.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
```diff
@@ -0,0 +1,109 @@
...
+        initialized = false;
+        canWrite = false;
+    }
+
+    private synchronized void maybeSetupPaths() throws IOException
+    {
+        if (initialized)
+            return;
+
+        File directory = mergedConfigDirectory.toFile();
```
hashbrowncipher, line 46: I'm from the church of EAFP, here to spread the good word.

Have you considered writing heavenly code, like so?

```
if(directory.mkdirs()) {
    Files.setPosixFilePermissions(mergedConfigDirectory, 0700);
}
```

Doing so would allow you to eliminate the canWrite variable, because you'll throw an exception instead. You could maybe even run this on every iteration (once per minute), as a single mkdir* syscall is not very resource-intensive.

    * I checked, and the JVM actually LBYLs it into a stat() syscall instead.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
```diff
@@ -0,0 +1,109 @@
...
+    public void execute() throws Exception
+    {
+        if (!initialized)
+            maybeSetupPaths();
+
+        Path structuredPathTmp = structuredPath.resolveSibling(structuredPath.getFileName() + ".tmp");
+        File output = structuredPathTmp.toFile();
+        output.createNewFile();
+        // The configuration might contain sensitive information, so ... don't let non Priam users read it
+        Files.setPosixFilePermissions(structuredPathTmp, PosixFilePermissions.fromString("rw-------"));
```
hashbrowncipher: I believe Files.createTempFile will do the work of opening a file with the desired permissions for you.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
```diff
@@ -0,0 +1,109 @@
...
+
+        Map<String, Object> structuredConfiguration = config.getStructuredConfiguration("all");
+
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectWriter structuredPathTmpWriter = mapper.writer(new MinimalPrettyPrinter());
+        structuredPathTmpWriter.writeValue(output, structuredConfiguration);
+
+        // Atomically swap out the new config for the old config.
+        if (!structuredPathTmp.toFile().renameTo(structuredPath.toFile()))
+            logger.warn("Failed to persist structured Priam configuration");
```
hashbrowncipher: nbd, but isn't this a failure of this thread to perform its core function? Should it be logger.error?

hashbrowncipher (approved):
Looks great. I made three stylistic notes, but my main priority is to just get something useful shipped.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
jolynch: Sure, simplified.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
jolynch: Hm, I didn't want to use that in case there was a bug and I created thousands of forgotten tmp files (like the last time that happened in Priam) ... But sure it's shorter

I don't want to implicitly rely on a non public guarantee of createTempFile though (afaik it doesn't guarantee the permissions). I'm going to leave the chmod just in case.

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
jolynch: Changed

priam/src/main/java/com/netflix/priam/config/PriamConfigurationPersister.java
jolynch: I ended up doing this just if the rename fails I delete it... hopefully there is no exception

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -795,4 +802,56 @@ default int getPostRestoreHookHeartbeatCheckFrequencyInMs() {
...
+    /**
+     * Cron expression to be used for persisting Priam merged configuration to disk. Use "-1" to disable the CRON.
+     * This will persist the fully merged value of Priam's configuration to the {@link #getMergedConfigurationDirectory()}
+     * as two JSON files: structured.json and unstructured.json which persist structured config and
+     * unstructured config respectively. We recommend you only rely on unstructured for the time being until the
+     * structured interface is finalized.
+     *
+     * Default: every minute
+     *
+     * @return Cron expression for flush
```
arunagrawal84: nit: typo - flush..

jolynch: Fixed

PR #732, opened 2018-10-08
Move forgotten files to lost_found directory 

arunagrawal84:
* Configuration: `priam.forgotten.file.move.enable`. Default: false. 
This will create the 'lost_found' directory in the column family and move any files Priam finds to be forgotten by Cassandra (and is at least a day old).

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
+++ b/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
@@ -249,7 +249,7 @@ public class SnapshotBackup extends AbstractBackup {
             final Path destDir = Paths.get(columnfamilyDir.getAbsolutePath(), "lost+found");
             for (File file : columnfamilyFiles) {
                 logger.warn("Forgotten file: {} found for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName());
-                if (config.shouldMoveForgottenFiles()) {
+                if (config.isForgottenFileMoveEnabled()) {
                     try {
                         FileUtils.moveFileToDirectory(file, destDir.toFile(), true);
                     } catch (IOException e) {
--- a/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
+++ b/priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
@@ -38,6 +38,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.*;
@@ -126,6 +128,7 @@ public class SnapshotBackup extends AbstractBackup {
 
         } catch (Exception e) {
             logger.error("Exception occurred while taking snapshot: {}. Exception: {}", snapshotName, e.getLocalizedMessage());
+            e.printStackTrace();
             snapshotStatusMgr.failed(backupMetadata);
             throw e;
         } finally {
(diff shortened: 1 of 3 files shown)


priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
```diff
@@ -1095,4 +1094,14 @@ public String getMergedConfigurationCronExpression() {
         // Every minute on the top of the minute.
         return config.get(PRIAM_PRE + ".configMerge.cron", "0 * * * * ? *");
     }
+
+    @Override
+    public int getForgottenFileGracePeriodDays() {
+        return config.get(PRIAM_PRE + ".forgotten.file.grace.period", 1);
```
vinaykumarchella: This is an opportunity to stop the bleeding for this naming standard. I would suggest using the bean naming standards that archaius uses.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -238,11 +241,17 @@ private void findForgottenFiles(File snapshotDir) {
             if (columnfamilyFiles.size() == 0)
                 return;
 
-            columnfamilyFiles.parallelStream().forEach(file -> logger.info("Forgotten file: {} found for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName()));
-
-            //TODO: The eventual plan is to move the forgotten files to a lost+found directory and clean the directory after 'x' amount of time. This behavior should be configurable. 
-            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
             logger.warn("# of forgotten files: {} found for CF: {}", columnfamilyFiles.size(), columnfamilyDir.getName());
```
vinaykumarchella, line 245: You might want to consider renaming this method since it not only finds the forgotten files but moves them too.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -238,11 +241,17 @@ private void findForgottenFiles(File snapshotDir) {
...
+            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
+
+            //Move the files to lost_found directory if configured.
+            final Path destDir = Paths.get(columnfamilyDir.getAbsolutePath(), "lost_found");
+            for (File file : columnfamilyFiles) {
+                logger.info("Forgotten file: {} found for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName());
+                if (config.shouldMoveForgottenFiles())
+                    FileUtils.moveFileToDirectory(file, destDir.toFile(), true);
```
vinaykumarchella: How do you handle exceptions while moving one of the files? Assume there is 1 file in the listing, due to some repeatable exceptions, this loop fails to move that 1 bad file, as a result there wont be any progress on the remaining files from the listing. Don't you want to record the exception but do the job for remaining files in the listing?

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -238,11 +241,17 @@ private void findForgottenFiles(File snapshotDir) {
...
-            //TODO: The eventual plan is to move the forgotten files to a lost+found directory and clean the directory after 'x' amount of time. This behavior should be configurable. 
-            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
             logger.warn("# of forgotten files: {} found for CF: {}", columnfamilyFiles.size(), columnfamilyDir.getName());
+            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
+
+            //Move the files to lost_found directory if configured.
+            final Path destDir = Paths.get(columnfamilyDir.getAbsolutePath(), "lost_found");
+            for (File file : columnfamilyFiles) {
+                logger.info("Forgotten file: {} found for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName());
+                if (config.shouldMoveForgottenFiles())
```
vinaykumarchella: You may consider moving this check above the for loop

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -238,11 +241,17 @@ private void findForgottenFiles(File snapshotDir) {
...
 
-            columnfamilyFiles.parallelStream().forEach(file -> logger.info("Forgotten file: {} found for CF: {}", file.getAbsolutePath(), columnfamilyDir.getName()));
-
-            //TODO: The eventual plan is to move the forgotten files to a lost+found directory and clean the directory after 'x' amount of time. This behavior should be configurable. 
-            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
             logger.warn("# of forgotten files: {} found for CF: {}", columnfamilyFiles.size(), columnfamilyDir.getName());
+            backupMetrics.incrementForgottenFiles(columnfamilyFiles.size());
+
+            //Move the files to lost_found directory if configured.
+            final Path destDir = Paths.get(columnfamilyDir.getAbsolutePath(), "lost_found");
```
vinaykumarchella: Will there be any conflict if the CF name in `lost_found` and created pre 2.1?

vinaykumarchella (commented):
Considering the forgotten file detection is working fine, this LGTM except few comments mentioned inline.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
arunagrawal84: I want to log messages even if you decide not to move the files to `lost+found` directory

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
arunagrawal84: Moved it to name "lost+found" as it is standard and C* does not like '+'

PR #737, opened 2018-10-12
Move to google java style.

arunagrawal84:


priam-cass-extensions/src/main/java/com/netflix/priam/cassandra/extensions/DataFetcher.java
```diff
@@ -46,29 +42,19 @@ public static String fetchData(String url)
             ByteArrayOutputStream bos = new ByteArrayOutputStream();
             responseStream = new DataInputStream((FilterInputStream) conn.getContent());
             int c = 0;
-            while ((c = responseStream.read(b, 0, b.length)) != -1)
-                bos.write(b, 0, c);
+            while ((c = responseStream.read(b, 0, b.length)) != -1) bos.write(b, 0, c);
```
vinaykumarchella, line 45: I am not a fan of this style, but I don't know other plugins which can do

```
while()
{
}
```

PR #738, opened 2018-10-17
Fix NPE while traversing the filesystem.

arunagrawal84:
Fix some typos and access of variables/functions.


priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -117,18 +120,23 @@ protected final void initiateBackup(
...
-        if (!dataDir.exists()) {
+        if (!dataDir.exists() || !dataDir.isDirectory()) {
             throw new IllegalArgumentException(
-                    "The configured 'data file location' does not exist: "
+                    "The configured 'data file location' does not exist or is not a directory: "
                             + config.getDataFileLocation());
         }
         logger.debug("Scanning for backup in: {}", dataDir.getAbsolutePath());
-        for (File keyspaceDir : dataDir.listFiles()) {
+        File[] keyspaceDirectories = dataDir.listFiles();
```
arunagrawal84, line 129: listFiles returns an array. Iterating over an array without copying it, returns NPE. Ideally, we should be using DirectoryStream when we wish to iterate over results. I will open another PR to remove all the calls to listFiles() where we directly iterate.

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -99,6 +103,21 @@ public void execute() throws Exception {
             sleeper.sleep(WAIT_TIME_MS);
         }
 
+        // Do not allow more than one snapshot to run at the same time. This is possible as this
+        // happens on CRON.
+        if (!lock.tryLock()) {
```
arunagrawal84, line 108: You could execute multiple calls to Snapshot and since this is singleton "snapshotName" (local variable) changes in the middle. The previous snapshot will then never be processed. Ideally, we should not be executing another snapshot if the previous one is still RUNNING.

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -108,6 +111,13 @@ public void execute() throws Exception {
             return;
         }
 
+        // Do not allow more than one snapshotMetaService to run at the same time. This is possible
+        // as this happens on CRON.
+        if (!lock.tryLock()) {
```
arunagrawal84, line 116: Avoid multiple RUNNING SnapshotMetaService calls. Does not make any sense.

vinaykumarchella (approved):
LGTM

priam/src/main/java/com/netflix/priam/backup/SnapshotBackup.java
```diff
@@ -99,6 +103,21 @@ public void execute() throws Exception {
...
+            logger.error("Snapshot Operation is already running! Try again later.");
```
sumanth-pasupuleti: nit: maybe a warn, since this is not necessary caused by application, but rather due to crons?

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -108,6 +111,13 @@ public void execute() throws Exception {
...
+            logger.error("SnapshotMetaService is already running! Try again later.");
```
sumanth-pasupuleti: similar nit on warn logging

sumanth-pasupuleti (approved):
couple of nits.

PR #757, opened 2018-11-25
Enable uploads from SnapshotMetaService in new backup format

arunagrawal84:
This allows snapshot files to be uploaded from backup version 2.0 in the new format. All the new files are uploaded and the backup file type used is SST_V2.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
@@ -65,23 +65,24 @@ public class S3BackupPath extends AbstractBackupPath {
 
     /*
      * This method will generate the location for the V2 backups.
-     * Note that we use epochMillis to sort the directory instead of traditional YYYYMMddHHmm. This will allow greater flexibility when doing restores
-     * as the s3 list calls with common prefix will have greater chance of match instead of traditional way (where it fails miserably when month or year changes).
-     * Also, to note that with backup v2 file layout, S3 list calls will be directly proportional to no. of tables in the cassandra clusters.
-     * For shared clusters (or clusters with huge no. of tables) this might be an issue and slowdowns in restore can happen, but the time will be always proportional vs. unpredictable time in V1 backups.
-     * The new way of storage will also allow for greater flexibility where we would be able to provide per table TTL for backups, if we want to in future.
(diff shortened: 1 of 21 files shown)


priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
```diff
@@ -32,54 +36,138 @@ public S3BackupPath(IConfiguration config, InstanceIdentity factory) {
         super(config, factory);
     }
 
+    private Path getV2Prefix() {
+        return Paths.get(baseDir, getAppNameWithHash(), instanceIdentity.getInstance().getToken());
+    }
+
+    private void parseV2Prefix(Path remoteFilePath) {
+        assert remoteFilePath.getNameCount() >= 3;
```
sumanth-pasupuleti: do you expect the name count to be > 3 as well?

priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
```diff
@@ -32,54 +36,138 @@ public S3BackupPath(IConfiguration config, InstanceIdentity factory) {
...
+        baseDir = remoteFilePath.getName(0).toString();
+        clusterName = parseAndValidateAppNameWithHash(remoteFilePath.getName(1).toString());
+        token = remoteFilePath.getName(2).toString();
+    }
+
+    /* This will ensure that there is some randomness in the path at the start so that remote file systems
+    can hash the contents better when we have lot of clusters backing up at the same remote location.
+    */
+    private String getAppNameWithHash() {
+        return (config.getAppName().hashCode() % 10000) + "_" + config.getAppName();
```
sumanth-pasupuleti: nit: I prefer using String.format

sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/aws/S3BackupPath.java
arunagrawal84: Yes, as getNameCount() just parses the no. of elements in the Path. These names are generated via getV2Prefix(). This check is to ensure that when we fail-fast if we can't call getName(0..2). Also, this method is used for parsing V2 in general (it is a sub-method), so it can be more than 3 e.g. when giving full path to META_V2 or SST_V2. 
 https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#getNameCount()

PR #758, opened 2018-12-02
Refactor S3 and GCS code for Iterators

arunagrawal84:
* Migrate Backup 1.0 Paths for more natural migration.
* Add utility class `Daterange` for easier input from operators. 
* Refactor S3 and GCS code, so individual file system only need to implement generic file iterator


priam/src/main/java/com/netflix/priam/backup/IBackupFileSystem.java
```diff
@@ -114,6 +114,16 @@ void uploadFile(
             final boolean deleteAfterSuccessfulUpload)
             throws FileNotFoundException, RejectedExecutionException, BackupRestoreException;
 
+    /**
+     * Get the bucket where this object should be stored. For local file system this should be empty
+     * or null.
+     *
+     * @return the location of the bucket.
+     */
+    default String getBucket() {
```
jolynch, line 123: Is this leaking the implementation details (S3) into the interface? What if it's just `getShard` or something?

jolynch (approved):
Shipit, lg2m

PR #759, opened 2018-12-08
Backup 2.0 Validation

arunagrawal84:


priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
```diff
@@ -90,27 +91,37 @@ private Path getV2Location() {
...
+                fileName);
     }
 
     private void parseV2Location(String remoteFile) {
         Path remoteFilePath = Paths.get(remoteFile);
         parseV2Prefix(remoteFilePath);
-        assert remoteFilePath.getNameCount() >= 6;
-        type = BackupFileType.valueOf(remoteFilePath.getName(3).toString());
-        setLastModified(Instant.ofEpochMilli(Long.parseLong(remoteFilePath.getName(4).toString())));
+        assert remoteFilePath.getNameCount() >= 8;
```
vinaykumarchella: Our prod and test fleet Priams do not run with assertions enabled, so having this assertion would not help, either enable assertions ( which I am not big fan of enabling it in prod) or validations would help

priam/src/main/java/com/netflix/priam/services/BackupTTLService.java
```diff
@@ -0,0 +1,222 @@
...
+import java.nio.file.Paths;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.*;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Created by aagrawal on 11/26/18. */
```
vinaykumarchella: Might be a good idea to document the goal of this class

vinaykumarchella:
LGTM, except few minor nits.

PR #762, opened 2018-12-26
Proxy layer for managing backup tasks

arunagrawal84:
Removal of deprecated and unused code.


priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -14,150 +14,60 @@
...
-        // Download meta.json from backup location and uncompress it.
-        List<String> metaFileList = new ArrayList<>();
-        try {
-            Path metaFileLocation =
-                    FileSystems.getDefault().getPath(config.getDataFileLocation(), "tmp_meta.json");
-            bkpStatusFs.downloadFile(Paths.get(metas.get(0).getRemotePath()), metaFileLocation, 5);
-            logger.info(
-                    "Meta file successfully downloaded to localhost: {}",
-                    metaFileLocation.toString());
+        metadata.sort((o1, o2) -> o2.getStart().compareTo(o1.getStart()));
```
vinaykumarchella: You could just include this in you stream above. something like this
`  metadata.stream()
                        .filter(backupMetadata -> backupMetadata.getStatus() == Status.FINISHED)
                        .sorted((o1, o2) -> o2.getStart().compareTo(o1.getStart()))
                        .collect(Collectors.toList());`

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -14,150 +14,60 @@
...
+        return metadata.stream().findFirst();
```
vinaykumarchella: I would be careful about findFirst(), since it can throw NPE. Either provide NullableOptional or nulls to your callers.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
vinaykumarchella: Also, do you want to consider which completed latest than which started first in the list of metadata files?

vinaykumarchella (commented):
Overall it is clean and well documented. Few minor comments.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
arunagrawal84: Can you clarify more? 

From the documentation of findFirst, you always get Optional and not NPE. Also, this method returns Optional<BackupMetaData> to ensure the same. 
```
 /**
     * Returns an {@link Optional} describing the first element of this stream,
     * or an empty {@code Optional} if the stream is empty.  If the stream has
     * no encounter order, then any element may be returned.
     *
     * <p>This is a <a href="package-summary.html#StreamOps">short-circuiting
     * terminal operation</a>.
     *
     * @return an {@code Optional} describing the first element of this stream,
     * or an empty {@code Optional} if the stream is empty
     * @throws NullPointerException if the element selected is null
     */
    Optional<T> findFirst();
```

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
arunagrawal84: It was sorted reverse w.r.t start date (made the change to mention that explicitly). Since we have a lock on the snapshot job, one which starts last (and is successful) will be one which finishes last as well.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
vinaykumarchella: if your stream produces null, then you would get NPE

PR #770, opened 2019-01-24
Add Cass SNAPSHOT JMX status, snapshot version, last validated timestamp. Changes to Servlet API.

arunagrawal84:


priam/src/main/java/com/netflix/priam/backup/BackupMetadata.java
```diff
@@ -29,19 +29,27 @@
     private String token;
     private Date start, completed;
     private Status status;
+    private boolean cassandraSnapshotSuccess;
+    private Date lastValidated;
+    private int backupVersion;
     private String snapshotLocation;
 
-    public BackupMetadata(String token, Date start) throws Exception {
+    public BackupMetadata(int backupVersion, String token, Date start) throws Exception {
```
jolynch: Can `backupVersion` be an ENUM?

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
```diff
@@ -174,6 +186,55 @@ public void failed(BackupMetadata backupMetadata) {
      */
     protected abstract LinkedList<BackupMetadata> fetch(String snapshotDate);
 
+    public List<BackupMetadata> getLatestBackupMetadata(
+            int snapshotVersion, DateUtil.DateRange dateRange) {
+        Instant startTime = dateRange.getStartTime().truncatedTo(ChronoUnit.DAYS);
```
jolynch: nit: I'd call these `startDay` and `endDay` to make it clear they're truncated to the latest day.

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
```diff
@@ -174,6 +186,55 @@ public void failed(BackupMetadata backupMetadata) {
...
+        Instant endTime = dateRange.getEndTime().truncatedTo(ChronoUnit.DAYS);
+
+        List<BackupMetadata> allBackups = new ArrayList<>();
+        Instant previousDay = endTime;
+        do {
+            // We need to find the latest backupmetadata in this date range.
+            logger.info(
+                    "Will try to find snapshot for : {}",
+                    DateUtil.formatInstant(DateUtil.yyyyMMddHHmm, previousDay));
+            List<BackupMetadata> backupsForDate = locate(new Date(previousDay.toEpochMilli()));
```
jolynch, line 201: Where does `locate` actually do S3 listing? From what I can tell it delegates to the in memory map, but I can't see where the map is populated?

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
```diff
@@ -174,6 +186,55 @@ public void failed(BackupMetadata backupMetadata) {
...
+        // the snapshot.
+        return allBackups
+                .stream()
+                .filter(backupMetadata -> backupMetadata != null)
+                .filter(backupMetadata -> backupMetadata.getStatus() == Status.FINISHED)
+                .filter(backupMetadata -> backupMetadata.getBackupVersion() == snapshotVersion)
+                .filter(
+                        backupMetadata ->
+                                backupMetadata
+                                                .getStart()
```
jolynch: nit: this is slightly odd to me, I feel like the `.getStart()` should happen on the same line as `backupMetadata`

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
```diff
@@ -174,6 +186,55 @@ public void failed(BackupMetadata backupMetadata) {
...
+                                                .toInstant()
+                                                .isAfter(dateRange.getStartTime())
+                                        || backupMetadata
```
jolynch: You can simplify this by just using `Instant.compareTo` and then you can also remove the second filter (change the `||` to an `&&` and use the two one liners for the same instant.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -33,39 +37,84 @@
...
+            throw new IllegalArgumentException("dateRange provided is null");
+        }
 
-        if (!latestBackupMetaData.isPresent()) {
-            logger.error("No backup found which finished during the time provided.");
-            return Optional.empty();
+        List<BackupMetadata> metadata =
+                backupStatusMgr.getLatestBackupMetadata(backupVersion, dateRange);
+        if (metadata == null || metadata.isEmpty()) return Optional.empty();
+        for (BackupMetadata backupMetadata : metadata) {
```
jolynch, line 83: So for backup 2.0, do we get the guarantee that old snapshots will eventually get all their data because we enumerate all snapshots and try to upload them regardless of when they happened (resumable backup)?

priam/src/main/java/com/netflix/priam/resources/BackupServlet.java
```diff
@@ -135,43 +131,24 @@ public Response status() throws Exception {
...
-            } else {
-                object.put("completetime", "not_available");
-            }
-
-        } else { // Backup do not exist for that date.
+        Optional<BackupMetadata> backupMetadataOptional =
+                this.completedBkups
+                        .locate(date)
+                        .stream()
+                        .filter(backupMetadata -> backupMetadata != null)
```
jolynch: nit: Can just use `Objects::nonNull`

priam/src/main/java/com/netflix/priam/resources/BackupServlet.java
```diff
@@ -135,43 +131,24 @@ public Response status() throws Exception {
...
-                            "http://localhost:8080/Priam/REST/v1/cassconfig/get_token");
-            if (token != null && !token.isEmpty()) {
-                object.put("token", token);
-            } else {
-                object.put("token", "not available");
-            }
-        }
+        } else {
 
+            object.put("Snapshotstatus", true);
```
jolynch, line 151: Did we not rely on the `token`, `start_time`, `complete_time` etc? 

If we are breaking backwards compatibility using snake case in json is really nice (e.g. `snapshot_status`, `details` etc ...)

priam/src/main/java/com/netflix/priam/resources/BackupServletV2.java
```diff
@@ -0,0 +1,113 @@
...
+    @Path("/ttl")
+    public Response ttl() throws Exception {
+        backupTTLService.execute();
+        return Response.ok(REST_SUCCESS, MediaType.APPLICATION_JSON).build();
+    }
+
+    @GET
+    @Path("/info/{date}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response info(@PathParam("date") String date) throws Exception {
```
jolynch: nix the throw

priam/src/main/java/com/netflix/priam/resources/BackupServletV2.java
```diff
@@ -0,0 +1,113 @@
...
+        Instant instant = DateUtil.parseInstant(date);
+        List<BackupMetadata> metadataList =
+                backupStatusMgr.getLatestBackupMetadata(
+                        SnapshotMetaService.BACKUP_VERSION,
+                        new DateRange(
+                                instant,
+                                instant.plus(1, ChronoUnit.DAYS).truncatedTo(ChronoUnit.DAYS)));
+        return Response.ok(metadataList.toString(), MediaType.APPLICATION_JSON).build();
```
jolynch: Do we want the `toString`? Without it we'll get a nice list of objects instead of a string that contains a serialized object right?

priam/src/main/java/com/netflix/priam/resources/BackupServletV2.java
```diff
@@ -0,0 +1,113 @@
...
+        Optional<BackupVerificationResult> result =
+                backupVerification.verifyBackup(
+                        SnapshotMetaService.BACKUP_VERSION, force, dateRange);
+        if (!result.isPresent()) {
+            return Response.noContent()
+                    .entity("No valid meta found for provided time range")
+                    .build();
+        }
+
+        return Response.ok(result.get().toString()).build();
```
jolynch: Same here re the `toString` ^^

priam/src/main/java/com/netflix/priam/services/BackupTTLService.java
```diff
@@ -87,7 +87,10 @@ public BackupTTLService(
     @Override
     public void execute() throws Exception {
         // Ensure that backup version 2.0 is actually enabled.
-        if (!backupRestoreConfig.enableV2Backups()) {
+        if (!backupRestoreConfig.enableV2Backups()
+                && backupRestoreConfig
+                        .getSnapshotMetaServiceCronExpression()
```
jolynch, line 92: If this is for the rollout period can we leave a comment indicating that?

priam/src/main/java/com/netflix/priam/services/SnapshotMetaService.java
```diff
@@ -68,6 +69,9 @@
     private final CassandraOperations cassandraOperations;
     private String snapshotName = null;
     private static final Lock lock = new ReentrantLock();
+    public static final int BACKUP_VERSION = 2;
```
jolynch: Still think this should be an enum ;-)

jolynch (approved):
Didn't look at the tests yet, but overall just some minor fix n ships and clarification questions.

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
arunagrawal84: It is left to interface with method `fetch`. A concrete implementation is available in FileSnapshotStatusMgr. Internally we fetch from DB.

priam/src/main/java/com/netflix/priam/backup/BackupStatusMgr.java
arunagrawal84: This is google java format, formatting the code.

priam/src/main/java/com/netflix/priam/services/BackupTTLService.java
arunagrawal84: Irrespective of the rollout, no point running TTL service, if V2 backups are not enabled.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
arunagrawal84: Yes, you are correct. We try to upload all the files to achieve resume. We try to upload at the start of Priam as well to ensure that if backup 2.0 cron is set far apart, we still try to upload those files and not wait until next run.

priam/src/main/java/com/netflix/priam/resources/BackupServlet.java
arunagrawal84: no we do not rely on other fields but "Snapshotstatus". Thus, not breaking the compatibility by keeping that field only. All other fields were added without any documentation as well so no one should rely on that.

PR #777, opened 2019-02-08
Do not check existence of file if it is not SST_V2.

arunagrawal84:
Note: AmazonServiceException can be for either Amazon unable to process request or for slow down. In either case, it is better to try to upload.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/aws/S3FileSystemBase.java
+++ b/priam/src/main/java/com/netflix/priam/aws/S3FileSystemBase.java
@@ -13,6 +13,7 @@
  */
 package com.netflix.priam.aws;
 
+import com.amazonaws.AmazonClientException;
 import com.amazonaws.services.s3.AmazonS3;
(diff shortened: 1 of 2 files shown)


priam/src/main/java/com/netflix/priam/aws/S3FileSystemBase.java
```diff
@@ -168,18 +166,12 @@ public boolean doesRemoteFileExist(Path remotePath) throws BackupRestoreExceptio
...
             exists = s3Client.doesObjectExist(getShard(), remotePath.toString());
-        } catch (AmazonServiceException ase) {
-            // Amazon S3 rejected request
-            throw new BackupRestoreException(
-                    "AmazonServiceException while checking existence of object: "
-                            + remotePath
-                            + ". Error: "
-                            + ase.getMessage());
-        } catch (AmazonClientException ace) {
+        } catch (Exception ex) {
```
jolynch: Can we just catch the two exceptions instead of all?

jolynch (approved):
Fix n ship

PR #779, opened 2019-02-09
Put a cache in front of remote file system call of objectExist.

arunagrawal84:
Add a rate limiter to s3 objectExist so we don't get slow down from S3.


priam/src/main/java/com/netflix/priam/aws/S3FileSystemBase.java
```diff
@@ -66,6 +67,10 @@
 
         double throttleLimit = config.getUploadThrottle();
         this.rateLimiter = RateLimiter.create(throttleLimit < 1 ? Double.MAX_VALUE : throttleLimit);
+
+        int objectExistLimit = config.getRemoteFileSystemObjectThrottle();
```
vinaykumarchella: You might want to consider changing dynamically, in case of incident?

vinaykumarchella (commented):
LGTM, fixNship

PR #789, opened 2019-02-22
Fix for ForgottenFiles when there are long running compaction job or Cassandra still reading from the files.

arunagrawal84:


priam/src/main/java/com/netflix/priam/backupv2/ForgottenFilesManager.java
```diff
@@ -125,22 +131,74 @@ public void findAndMoveForgottenFiles(Instant snapshotInstant, File snapshotDir)
...
+            }
+        }
+
+        // Clean LOST_FOUND directory of any previous symbolic link files which are not considered
+        // lost any more.
+        for (File file : FileUtils.listFiles(destDir.toFile(), null, false)) {
+            Path filePath = Paths.get(file.getAbsolutePath());
+            if (Files.isSymbolicLink(filePath)) {
+                Path originalFile = Files.readSymbolicLink(filePath);
+                if (!columnfamilyPaths.contains(originalFile)) Files.delete(filePath);
```
vinaykumarchella: Might be a good idea to log a message here, this also indicates how many times we got into above mentioned C* edgecases.

priam/src/main/java/com/netflix/priam/backupv2/ForgottenFilesManager.java
```diff
@@ -125,22 +131,74 @@ public void findAndMoveForgottenFiles(Instant snapshotInstant, File snapshotDir)
...
+                    Instant last_modified_time =
+                            Files.getLastModifiedTime(symbolic_link, LinkOption.NOFOLLOW_LINKS)
+                                    .toInstant();
+                    if (DateUtil.getInstant()
+                            .isAfter(
+                                    last_modified_time.plus(
+                                            config.getForgottenFileGracePeriodDaysForRead(),
+                                            ChronoUnit.DAYS))) {
+                        // Eligible for move.
+                        logger.warn(
```
vinaykumarchella: Do you want to move this warn line to below  `config.isForgottenFileMoveEnabled()`. Currently, there is no way for us to track which files actually got moved to LOST_FOUND directory.

priam/src/main/java/com/netflix/priam/backupv2/ForgottenFilesManager.java
```diff
@@ -125,22 +131,74 @@ public void findAndMoveForgottenFiles(Instant snapshotInstant, File snapshotDir)
...
+                                        "Exception occurred while trying to move forgottenFile: {}. Ignoring the error and continuing with remaining backup/forgotten files.",
+                                        file);
+                                e.printStackTrace();
+                            }
+                        }
+                    }
                 }
+
+            } catch (IOException e) {
+                logger.error("Forgotten file: Error while trying to process the file: {}", file);
```
vinaykumarchella, line 195: Want to add the stacktrace too? To diagnose the issue

vinaykumarchella (commented):
First pass review comments inline. Except for a few minor nits, LGTM. I will take another look at it later today along with testcases

priam/src/main/java/com/netflix/priam/backupv2/ForgottenFilesManager.java
arunagrawal84: Wanted a log message to find forgotten files even if we don't move them. Moved to INFO level and added WARN statement after moving to ensure we capture the successfully moved files.

PR #792, opened 2019-02-27
S3 - BucketLifecycleConfiguration has `prefix` method removed from latest library. 

arunagrawal84:


priam/src/main/java/com/netflix/priam/aws/S3FileSystemBase.java
```diff
@@ -121,42 +129,77 @@ public void cleanup() {
...
     }
 
-    private boolean updateLifecycleRule(IConfiguration config, List<Rule> rules, String prefix) {
-        Rule rule = null;
-        for (BucketLifecycleConfiguration.Rule lcRule : rules) {
-            if (lcRule.getPrefix().equals(prefix)) {
-                rule = lcRule;
-                break;
+    // Dummy class to get Prefix. - Why oh why AWS you can't give the details!!
+    private class PrefixVisitor implements LifecyclePredicateVisitor {
```
jolynch, line 133: Think this should either be static or an anonymous class (something something memory leaks)

jolynch (approved):
Fix n ship

PR #795, opened 2019-03-01
Fix X->Y->Z issue. Replace nodes when gossip actually converges.

arunagrawal84:


priam/src/main/java/com/netflix/priam/defaultimpl/CassandraOperations.java
```diff
@@ -126,4 +127,73 @@ public Void retriableCall() throws Exception {
             }
         }.call();
     }
+
+    @Override
+    public List<Map<String, String>> gossipInfo() throws Exception {
```
vinaykumarchella: This parsing logic tends to break across the versions, can you add a test case by getting this info from local C* and ensure that parsing logic does not fail in the version that branch is supporting.

priam/src/main/java/com/netflix/priam/identity/token/DeadTokenRetriever.java
```diff
@@ -67,22 +63,6 @@ public DeadTokenRetriever(
 
         List<String> crossAccountAsgInstances = membership.getCrossAccountRacMembership();
 
-        if (logger.isInfoEnabled()) {
```
vinaykumarchella: Nit: Good idea to isolate the changes in future. Easy to cherry pick, revert etc.,

priam/src/main/java/com/netflix/priam/identity/token/DeadTokenRetriever.java
```diff
@@ -107,45 +88,84 @@ public PriamInstance get() throws Exception {
...
-            return factory.create(
-                    config.getAppName(),
-                    markAsDead.getId(),
-                    instanceInfo.getInstanceId(),
-                    instanceInfo.getHostname(),
-                    instanceInfo.getHostIP(),
-                    instanceInfo.getRac(),
-                    markAsDead.getVolumes(),
-                    payLoad);
+                    "Will try to replace token: {} with replacedIp (from gossip info): {} instead of ip from DB: {}",
```
vinaykumarchella: `from DB` --> `from turtle`?

priam/src/main/java/com/netflix/priam/identity/token/DeadTokenRetriever.java
```diff
@@ -154,87 +174,90 @@ public String getReplaceIp() {
...
+                    "Error in reaching out to host: [{}} or parsing response from host: {}",
+                    host,
+                    response);
         }
         return null;
     }
 
-    private String getBaseURI(String host) {
-        return "http://" + host + ":8080/";
+    private String getURI(String host) {
```
vinaykumarchella: I would be specific on what this method returns?

priam/src/main/java/com/netflix/priam/resources/CassandraAdmin.java
```diff
@@ -414,153 +419,9 @@ public Response statusthrift()
...
-                    obj.put(kv[0], kv[1]);
-                }
-            }
-        }
-        if (StringUtils.isNotBlank(key)) rootObj.put(key, obj);
-        return rootObj;
-    }
-
-    @GET
-    @Path("/netstats")
```
vinaykumarchella: same as above, it would be cleaner to separate these commits.

priam/src/main/java/com/netflix/priam/identity/token/DeadTokenRetriever.java
arunagrawal84: turtle is Netflix token DB (not common). Token DB, might be more appropriate.

vinaykumarchella (approved):

PR #799, opened 2019-03-06
Send notification only if we upload the file.

arunagrawal84:


vinaykumarchella (approved):
LGTM, however, surprised that read after write of S2 guarantee is not helping you here.

PR #801, opened 2019-03-09
Hotfix: TTL for backup version 2.

arunagrawal84:
Cassandra can flush files on file system like Index.db first and other component files later (like 30 mins). If there is a snapshot in between, then this "single" component file would not be part of the snapshot as SSTable is still not part of Cassandra's "view". Only if Cassandra could provide strong guarantees on the file system such that  -
1. All component will be flushed to disk as real SSTables only if they are part of the view. Until that happens all the files will be "tmp" files.
2. All component flushed will have the same "last modified" file. i.e. on the first flush. Stats.db can change over time and that is OK.

Since this is not the case, the TTL may end up deleting this file even though the file is part of the next snapshot. To avoid, this we add grace period (based on how long compaction can run) when we delete the files.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
+++ b/priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
@@ -37,6 +37,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.*;
 import org.apache.commons.collections4.iterators.FilterIterator;
@@ -62,6 +63,9 @@ public abstract class AbstractFileSystem implements IBackupFileSystem, EventGene
     private final Set<Path> tasksQueued;
     private final ThreadPoolExecutor fileUploadExecutor;
     private final ThreadPoolExecutor fileDownloadExecutor;
+
+    // This is going to be a write-thru cache containing the most frequently used items from remote
+    // file system. This is to ensure that we don't make too many API calls to remote file system.
     private final Cache<Path, Boolean> objectCache;
 
     @Inject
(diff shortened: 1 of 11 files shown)


priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
```diff
@@ -722,7 +722,7 @@ public String getMergedConfigurationCronExpression() {
     }
 
     @Override
-    public int getForgottenFileGracePeriodDaysForCompaction() {
+    public int getGracePeriodDaysForCompaction() {
```
vinaykumarchella, line 725: Do you want to change `PRIAM_PRE + ".forgottenFileGracePeriodDaysForCompaction` also to `gracePeriodDaysForCompaction`

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -257,6 +258,16 @@ public boolean checkObjectExists(Path remotePath) {
         return remoteFileExist;
     }
 
+    @Override
+    public void deleteRemoteFiles(List<Path> remotePaths) throws BackupRestoreException {
+        if (remotePaths == null) return;
+
+        deleteFiles(remotePaths);
+        objectCache.invalidateAll(remotePaths);
```
vinaykumarchella, line 273: I would invalidate cache before deleting remote files. Also, code documentation on why this is done this way.

priam/src/main/java/com/netflix/priam/services/BackupTTLService.java
```diff
@@ -130,15 +126,45 @@ public void execute() throws Exception {
...
+                        metas.get(metas.size() - 1).getLastModified());
+                for (AbstractBackupPath meta : metas) {
+                    deleteFile(meta, false);
+                }
+            }
+
             Iterator<String> remoteFileLocations =
                     fileSystem.listFileSystem(getSSTPrefix(), null, null);
 
+            dateToTtl = dateToTtl.minus(config.getGracePeriodDaysForCompaction(), ChronoUnit.DAYS);
```
vinaykumarchella: Context on why you are considering `getGracePeriodDaysForCompaction` instead of TTL only.

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
arunagrawal84: Ack.

priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
arunagrawal84: Ack

priam/src/main/java/com/netflix/priam/services/BackupTTLService.java
arunagrawal84: Ack

PR #814, opened 2019-04-27
Priam will check gossip info while grabbing pre-assigned token.

cthumuluru:
Priam will check Cassandra gossip information while grabbing pre-assigned token to decide if it should start Cassandra in bootstrap mode or in replace mode. Moved token owner inferring logic based on Cassandra gossip into a util class. Refactored InstanceIdentity.init() method.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
+++ b/priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
@@ -28,13 +28,18 @@ import com.netflix.priam.identity.config.InstanceInfo;
 import com.netflix.priam.identity.token.IDeadTokenRetriever;
 import com.netflix.priam.identity.token.INewTokenRetriever;
 import com.netflix.priam.identity.token.IPreGeneratedTokenRetriever;
+import com.netflix.priam.identity.token.TokenRetrieverUtils;
+import com.netflix.priam.identity.token.TokenRetrieverUtils.GossipParseException;
 import com.netflix.priam.utils.ITokenManager;
 import com.netflix.priam.utils.RetryableCallable;
 import com.netflix.priam.utils.Sleeper;
 import java.net.UnknownHostException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -74,8 +79,12 @@ public class InstanceIdentity {
     private final IPreGeneratedTokenRetriever preGeneratedTokenRetriever;
     private final INewTokenRetriever newTokenRetriever;
 
+    private final java.util.function.Predicate<PriamInstance> sameHostPredicate =
+            (i) -> i.getInstanceId().equals(myInstanceInfo.getInstanceId());
+
     @Inject
-    // Note: do not parameterized the generic type variable to an implementation as it confuses
+    // Note: do not parameterized the generic type variable to an implementation as
+    // it confuses
     // Guice in the binding.
     public InstanceIdentity(
             IPriamInstanceFactory factory,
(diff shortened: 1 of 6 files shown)


arunagrawal84:
Can you add a test case to cover the scenario where priam has already grabbed the token but C* do not agree and we use replace ip.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
```diff
@@ -109,154 +117,162 @@ public InstanceInfo getInstanceInfo() {
...
-                                        != null) { // The IP address of the dead instance to which
-                                    // we will acquire its token
-                                    replacedIp = deadTokenRetriever.getReplaceIp();
-                                }
-                            }
-
-                            return result;
-                        }
+        // Grab the token which was already assigned.
+        myInstance = mayGrabAlreadyAssignedToken();
```
arunagrawal84: nit: grabPreAssignedToken();

cthumuluru:
Reverted the logic to get the replaceIP to only return the IP when all the desired nodes in the gossip agree on the IP to be replaced.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
```diff
@@ -109,154 +117,162 @@ public InstanceInfo getInstanceInfo() {
...
+
+    private PriamInstance mayGrabDeadToken() throws Exception {
+        return new RetryableCallable<PriamInstance>() {
+            @Override
+            public PriamInstance retriableCall() throws Exception {
+                PriamInstance result = deadTokenRetriever.get();
+                if (result != null) {
+                    isReplace = true; // indicate that we are acquiring a dead instance's token.
+
+                    if (deadTokenRetriever.getReplaceIp()
```
arunagrawal84, line 216: If for any reason, if we are not able to communicate to any other instance (say missed security group configuration, all other nodes are down etc), we should just believe the token DB. Currently, we will get "null" as replaceIp and we will never be able to bootstrap. We should do something - 
`if(deadTokenRetriever.getReplaceIp()==null) { //Believe tokenDB replaceIp = result.getIp(); // WARN message }`

cthumuluru:
added unit tests for both pre-assigned token retriever and token retriever utils.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
```diff
@@ -0,0 +1,122 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(TokenRetrieverBase.class);
+    private static final String GOSSIP_INFO_URL_FORMAT =
+            "http://%s:8080/Priam/REST/v1/cassadmin/gossipinfo";
+
+    /**
+     * Utility method to infer the IP of the owner of a token in a given datacenter. This method
+     * uses Cassandra gossip information to find the owner. The IP with maximum votes is inferred to
+     * be the token owner.
+     *
+     * @param allIds
```
arunagrawal84, line 29: Javadoc please.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
```diff
@@ -109,154 +117,162 @@ public InstanceInfo getInstanceInfo() {
...
 
-                        @Override
-                        public void forEachExecution() {
-                            populateRacMap();
-                            deadTokenRetriever.setLocMap(locMap);
-                        }
-                    }.call();
+        // Grab a dead token.
+        if (myInstance == null) {
+            myInstance = grabDeadToken();
```
arunagrawal84, line 128: Can you add INFO log statement when we move from one case to other? It is easier to debug in case of any failure.

cthumuluru: Yes, make sense. Done.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
cthumuluru: Updated the Javadoc to reflect the exact behavior.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
cthumuluru: This is taken care of in `grabAlreadyAssignedToken()` method. It returns the instance and C* is started in bootstrap mode.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
cthumuluru: done.

priam/src/main/java/com/netflix/priam/identity/InstanceIdentity.java
arunagrawal84: I do not agree with this statement. From infer token ownership in DeadToken we can return null. 
`logger.warn(
                "Return null: Unable to find enough instances where gossip match. Required: {}",
                noOfInstancesGossipShouldMatch);
        return null;`

In that case, DeadTokenRetriever will return null and thus no IP will be set in replacedIp. We overcome this particular thing in NFDeadTokenRetriever and my goal was to bring that goodness here as well with enough documentation.

cthumuluru: I modified TokenRetrievalUtils to throw GossipParseException when it is not able to fetch the gossip info or not able to parse the token info. With the exception we could differentiate gossip mismatch with gossip error. In case of gossip errror, the deadTokenRetriever falls back to the IP from token database.

priam/src/main/java/com/netflix/priam/identity/token/DeadTokenRetriever.java
```diff
@@ -118,19 +113,25 @@ public PriamInstance get() throws Exception {
...
+                if (replacedIp == null) return null;
+
+                logger.info(
+                        "Will try to replace token: {} with replacedIp (from gossip info): {} instead of ip from Token database: {}",
+                        priamInstance.getToken(),
+                        replacedIp,
+                        priamInstance.getHostIP());
+            } catch (GossipParseException e) {
+                // In case of gossip exception, fallback to IP in token database.
+                this.replacedIp = priamInstance.getHostIP();
```
arunagrawal84, line 133: LGTM

arunagrawal84 (approved):
LGTM

PR #821, opened 2019-05-25
Backport Configuring JVM Options for Cassandra using jvm.options

jolynch:
This backports the Priam functionality for tuning jvm.options files to
this branch. By default we don't use it, but if your version of
Cassandra 2.1 happens to have jvm.options support, we can use that here.

Commit message from 3.11 Priam branch:

Cassandra 3.x added a new way to configure heap sizes and pass other JVM
parameters (via jvm.options). Priam now supports configuring common
options like heap setting and choosing Garbage Collection type
(G1GC/CMS) natively. Default being CMS. It logs jvm.options after tuning
them.


arunagrawal84 (commented):
LGTM

PR #825, opened 2019-06-07
rolling back the changes to use gossip info while grabbing pre-assign…

cthumuluru:
Rolling back the changes to use gossip info while grabbing pre-assigned and dead tokens. Gossip information doesn't seem to reflect the correct state of the cluster always. This is blocking C* nodes from joining the ring. Temporarily rolling back the fix to use Gossip info.


jolynch (approved):
Looks like the right temporary fix, let's ship this hotfix asap. thanks for the fast fix!

PR #827, opened 2019-06-21
Removing Incremental manifest file functionality and bug fix. 

arunagrawal84:


sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
```diff
@@ -74,14 +69,20 @@ public static String inferTokenOwnerFromGossip(
...
+                    // We do not want to do anything if token is considered as alive by Cassandra.
+                    logger.info(
+                            "Not producing anything in replaceIp as according to C* that token is still alive or "
+                                    + "there is a mismatch in status information per Cassandra. ip: [{}], replaceIp: [{}]",
+                            ip,
+                            replaceIp);
+                    return null;
                 }
 
                 matchedGossipInstances++;
```
cthumuluru, line 88: `matchedGossipInstances` may not be needed anymore right? In case of a mismatch, we are returning null anyway.

cthumuluru (approved):
LGTM.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
arunagrawal84: We need that so that we come out of the loop fast. As soon as matchedGossipInstances == desired (3 in most cases) we want to exit the loop.

PR #833, opened 2019-09-20
Add support for compressions and make LZ4 default for backup 2.0 

arunagrawal84:
We have supported Snappy as the only compression type while uploading the files. We wanted to support customizable compression types and thus support for LZ4 and no compression. 
In this PR we make LZ4 as the default compression type for backup 2.0 
Future work: Do not compress Data.db which are already compressed. 

In this PR we have encapsulated the directives on how to upload or download a file to class UploadDownloadDirectives. This includes directives like compression, encryption, retries to use etc.


priam/src/main/java/com/netflix/priam/aws/S3FileSystem.java
```diff
@@ -213,7 +213,14 @@ protected long uploadFileImpl(Path localPath, Path remotePath) throws BackupRest
...
                 byte[] chunk = byteArrayOutputStream.toByteArray();
                 long compressedFileSize = chunk.length;
-                rateLimiter.acquire(chunk.length);
+
+                /**
+                 * Weird, right that we are checking for length which is positive. You can thanks
+                 * this to sometimes C* creating files which are zero bytes, and giving that in
+                 * snapshot for some unknown reason.
+                 */
+                if (chunk.length > 0) rateLimiter.acquire(chunk.length);
```
sumanth-pasupuleti, line 234: I assume we still want to proceed with the upload of this 0 byte file to keep our s3 snapshots consistent with C* snapshot view, and hence we do not skip the rest of the upload code path even if the file has 0 byte - is this correct?

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -245,16 +247,16 @@ public Long retriableCall() throws Exception {
...
 
     @Override
     public boolean checkObjectExists(Path remotePath) {
         // Check in cache, if remote file exists.
-        Boolean cacheResult = objectCache.getIfPresent(remotePath);
+        Instant lastUpdatedTimestamp = objectCache.getIfPresent(remotePath);
 
         // Cache hit. Return the value.
-        if (cacheResult != null) return cacheResult;
+        if (lastUpdatedTimestamp != null) return true;
```
sumanth-pasupuleti, line 271: since we do not really check the value of the timestamp, what difference does this make relative to having a Boolean? Maybe I will discover its importance in other fails, if I do, will delete this comment.

priam/src/test/java/com/netflix/priam/backup/s3test.java
```diff
@@ -0,0 +1,192 @@
...
+import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import com.amazonaws.services.s3.model.*;
+import java.io.*;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import javax.crypto.KeyGenerator;
+
+public class s3test {
```
sumanth-pasupuleti, line 33: This class has a lot of commented code. Any reason for keeping the commented code?

sumanth-pasupuleti (approved):

PR #837, opened 2019-11-12
bypass cassandra streaming

arunagrawal84:
Build the instance from backups by using the restore process in case of an instance replacement. Note that we prefer this when data size is HUGE. C* streaming is super slow, and for instances with big data size can lead to C* streaming for multiple days. Note that this is a little bit dangerous as you "will" lose writes accepted by the old instance but not uploaded to the backup file system. Also, we do not plan to run a local repair on the replaced instance, so data will be stale. We hope that the repair service will take care of the inconsistency. Clusters with LOCAL_QUORUM for reads and writes may see little to no impact.  If the restore fails, then we fall back to use "streaming".

### DIFF
--- a/priam/src/main/java/com/netflix/priam/PriamServer.java
+++ b/priam/src/main/java/com/netflix/priam/PriamServer.java
@@ -22,6 +22,7 @@ import com.netflix.priam.aws.UpdateSecuritySettings;
 import com.netflix.priam.backup.BackupService;
 import com.netflix.priam.backupv2.BackupV2Service;
 import com.netflix.priam.cluster.management.ClusterManagementService;
+import com.netflix.priam.config.IBackupRestoreConfig;
 import com.netflix.priam.config.IConfiguration;
 import com.netflix.priam.config.PriamConfigurationPersister;
 import com.netflix.priam.defaultimpl.ICassandraProcess;
@@ -42,6 +43,7 @@ import org.slf4j.LoggerFactory;
 public class PriamServer implements IService {
     private final PriamScheduler scheduler;
     private final IConfiguration config;
+    private final IBackupRestoreConfig backupRestoreConfig;
     private final InstanceIdentity instanceIdentity;
     private final Sleeper sleeper;
     private final ICassandraProcess cassProcess;
@@ -56,6 +58,7 @@ public class PriamServer implements IService {
     @Inject
     public PriamServer(
             IConfiguration config,
+            IBackupRestoreConfig backupRestoreConfig,
             PriamScheduler scheduler,
             InstanceIdentity id,
             Sleeper sleeper,
@@ -66,6 +69,7 @@ public class PriamServer implements IService {
             CassandraTunerService cassandraTunerService,
             ClusterManagementService clusterManagementService) {
         this.config = config;
+        this.backupRestoreConfig = backupRestoreConfig;
         this.scheduler = scheduler;
         this.instanceIdentity = id;
         this.sleeper = sleeper;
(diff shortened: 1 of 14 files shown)


priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -111,25 +115,48 @@ public void scheduleService() throws Exception {
...
+                    SystemUtils.cleanupDir(config.getDataFileLocation(), null);
+                } finally {
+                    shouldStartCassandra = true;
+                }
+            } else {
+                // no restores needed
+                logger.info("No restore needed, task not scheduled");
+                shouldStartCassandra = true;
+            }
         }
```
hashbrowncipher, line 151: How can we exit this block with shouldStartCassandra being false?

priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -111,25 +115,48 @@ public void scheduleService() throws Exception {
...
-            logger.info("No restore needed, task not scheduled");
-            if (!config.doesCassandraStartManually()) cassProcess.start(true); // Start cassandra.
-            else
-                logger.info(
-                        "config.doesCassandraStartManually() is set to True, hence Cassandra needs to be started manually ...");
+            // Start cassandra only if restore is successful.
+            shouldStartCassandra = true;
+        } else {
+            if (instanceIdentity.isReplace()
+                    && backupRestoreConfig.enableBypassCassandraStreaming()) {
```
hashbrowncipher, line 133: How does Priam determine whether Cassandra hasn't successfully bootstrapped? I'm looked for an existing check, but I didn't see one.

priam/src/main/java/com/netflix/priam/config/IBackupRestoreConfig.java
```diff
@@ -97,4 +97,18 @@ default int getBackupVerificationSLOInHours() {
     default boolean enableV2Restore() {
         return false;
     }
+
+    /**
+     * Build the instance from backups by using restore process in case of an instance replacements.
+     * Note that we prefer this when data size is HUGE. C* streaming is super slow and for instances
+     * with big data size can lead to C* streaming for multiple days. Note that this is a little bit
+     * dangerous as you "will" some of the writes accepted by old instance but not uploaded to
```
hashbrowncipher, line 105: I think a word is missing here.

priam/src/main/java/com/netflix/priam/config/IBackupRestoreConfig.java
```diff
@@ -97,4 +97,18 @@ default int getBackupVerificationSLOInHours() {
...
+     * backup file system. Also we do not plan to run local repair on the replaced instance, so data
```
hashbrowncipher, line 106: I agree that not running repair is acceptable for a first iteration. Hypothetically though, how would we do it?

priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
```diff
@@ -63,11 +67,13 @@ public void writeAllProperties(String yamlLocation, String hostname, String seed
         map.put("native_transport_port", config.getNativeTransportPort());
         map.put("listen_address", hostname);
         map.put("rpc_address", hostname);
+
         // Dont bootstrap in restore mode
-        if (!Restore.isRestoreEnabled(config, instanceInfo)) {
-            map.put("auto_bootstrap", config.getAutoBoostrap());
-        } else {
+        if (instanceState.getRestoreStatus() != null
```
hashbrowncipher, line 72: I don't see the purpose of this check. Why not just pass the `auto_bootstrap` setting through from the config 100% of the time?

hashbrowncipher (commented):
A few questions, at least some of them due to my unfamiliarity with Priam's codebase.

priam/src/main/java/com/netflix/priam/config/IBackupRestoreConfig.java
arunagrawal84: Ideally, we should be deferring that task to the repair service. Where that repair service sits, how it gets executed is a different conversation though.

priam/src/main/java/com/netflix/priam/PriamServer.java
arunagrawal84: TokenRetrieverUtils.inferTokenOwnerFromGossip is used to fetch the instance identity. That method should tell correctly if Cassandra had already bootstrapped successfully.

priam/src/main/java/com/netflix/priam/PriamServer.java
arunagrawal84: If the original requested `restore` was a failure (that is when Priam starts in restore mode during weekly restore refresh). `
`` `       // Start cassandra only if restore is successful.
            shouldStartCassandra = true;```

But yes with the recent refactoring of restore we will throw exception there and thus we don't need that variable. good catch.

priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
arunagrawal84: Idea is - if we are doing a restore then we need to set auto_bootstrap to `false` else use the provided value in the configuration provider (like when creating a new cluster with `false` or `true` for most of the cases). Since we can choose to restore to bypass Cassandra streaming we need to override the configured value. I wanted to keep that logic in tuner instead of putting in a configuration provider.

hashbrowncipher: Can't we leave `auto_bootstrap=true`, even when we restore?

priam/src/main/java/com/netflix/priam/PriamServer.java
```diff
@@ -111,25 +115,48 @@ public void scheduleService() throws Exception {
...
+                logger.info("Trying to download data instead of streaming from Cassandra.");
```
sumanth-pasupuleti, line 134: nit: Add "from backup", as in "Trying to download data instead of streaming from Cassandra"

sumanth-pasupuleti (approved):

PR #838, opened 2020-02-11
A first pass commit of adding filtering logic for enabling Priam to o…

schakrovorthy:
…nly send back up notifications for component files that are configured.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -1091,4 +1091,14 @@ public interface IConfiguration {
      * @return The result for the property, or the defaultValue if provided (null otherwise)
      */
     String getProperty(String key, String defaultValue);
+
+    /**
+     * Returns a csv of backup component file types {@link
+     * com.netflix.priam.backup.AbstractBackupPath.BackupFileType} on which to send notifications.
+     *
+     * @return A csv string that can be parsed to infer the component file types
+     */
+    default String getBackupNotifyComponentIncludeList() {
+        return StringUtils.EMPTY;
+    }
 }
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -1091,14 +1091,4 @@ public interface IConfiguration {
      * @return The result for the property, or the defaultValue if provided (null otherwise)
      */
     String getProperty(String key, String defaultValue);
-
-    /**
-     * Returns a csv of backup component file types {@link
-     * com.netflix.priam.backup.AbstractBackupPath.BackupFileType} on which to send notifications.
-     *
-     * @return A csv string that can be parsed to infer the component file types
-     */
-    default String getBackupNotifyComponentIncludeList() {
-        return StringUtils.EMPTY;
-    }
 }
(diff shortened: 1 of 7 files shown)


priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -1091,4 +1091,14 @@ default String getMergedConfigurationDirectory() {
...
      */
     String getProperty(String key, String defaultValue);
+
+    /**
+     * Returns a csv of backup component file types {@link
+     * com.netflix.priam.backup.AbstractBackupPath.BackupFileType} on which to send notifications.
+     *
+     * @return A csv string that can be parsed to infer the component file types
+     */
+    default String getBackupNotifyComponentIncludeList() {
```
arunagrawal84: Can you move this to BackupRestoreConfig? We are not adding any new backup restore config here.

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -1091,4 +1091,14 @@ default String getMergedConfigurationDirectory() {
      * @return The result for the property, or the defaultValue if provided (null otherwise)
      */
     String getProperty(String key, String defaultValue);
+
+    /**
+     * Returns a csv of backup component file types {@link
```
arunagrawal84: In the comment can you explain the default value of StringUtils.EMPTY? Also, provide an example of override value.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -51,6 +53,13 @@ public BackupNotificationMgr(
         this.notificationService = notificationService;
         this.instanceInfo = instanceInfo;
         this.instanceIdentity = instanceIdentity;
+        this.notifiedBackupFileTypesSet = new HashSet<>();
+        if (!StringUtils.isBlank(this.config.getBackupNotifyComponentIncludeList())) {
+            notifiedBackupFileTypesSet.addAll(
+                    Arrays.stream(this.config.getBackupNotifyComponentIncludeList().split(","))
+                            .map(AbstractBackupPath.BackupFileType::fromString)
```
arunagrawal84: You will get RuntimeException here if value contains illegal value. What does the error handling looks like?  I would recommend to reject the configuration change and keep StringUtils.EMPTY if it has illegal value. Also, adding javadoc for failure behavior in the configuration woud be a good idea. 

Also, this value will be updated only at startup. Can you make this dynamic so we don't need to restart Priam if we wish to change the value.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -97,27 +106,35 @@ public void notify(BackupVerificationResult backupVerificationResult) {
...
-            Map<String, MessageAttributeValue> messageAttributes =
-                    getMessageAttributes(abp.getType());
+                // SNS Attributes for filtering messages. Cluster name and backup file type.
+                Map<String, MessageAttributeValue> messageAttributes =
+                        getMessageAttributes(abp.getType());
 
-            this.notificationService.notify(jsonObject.toString(), messageAttributes);
+                this.notificationService.notify(jsonObject.toString(), messageAttributes);
+            } else {
+                logger.info(
```
arunagrawal84: Please make this debug statement else log will be flooded with these messages.

arunagrawal84 (commented):
Fix n ship

arunagrawal84:
Sorry, forgot to mention, can you please add a test case for the filter to validate it works. It seems code coverage is declining by 0.12%

schakrovorthy:
> Sorry, forgot to mention, can you please add a test case for the filter to validate it works. It seems code coverage is declining by 0.12%
Looking good now?

schakrovorthy:
@arunagrawal84 Could you please review this PR now after I addressed your review comments.

PR #846, opened 2020-04-01
Feature/verify all back ups in date range

schakrovorthy:


priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
```diff
@@ -88,22 +89,32 @@ public void execute() throws Exception {
...
-            backupNotificationMgr.notify(verificationResult.get());
+            verificationResults
+                    .stream()
+                    .forEach(
+                            r -> {
+                                logger.info(
+                                        "Sending {} message for backup: {}",
+                                        AbstractBackupPath.BackupFileType.SNAPSHOT_VERIFIED,
+                                        r.snapshotInstant);
+                                backupNotificationMgr.notify(r);
```
arunagrawal84: As per your method of verifyAllBackups we add all the backups to the list (even ones which were verified before). This will basically make the logic as - "send verifications for all the backups in the SLO window every time we run this job which is every hour". Thus we will end up sending 48 notifications every hour for every node. It does get the job done but we are sending too many notifications which dont need to be sent. 
We should send notifications for all the "newly" verified snapshots and the latest snapshot ONLY. Most of the times we will send notification for "latest" snapshot with sometimes sending past message.

schakrovorthy: Makes sense. Let me see how I can optimize the number of verifications and thereby notifications.

PR #848, opened 2020-04-22
Addressing review comments

schakrovorthy:
This PR is created to address the review comment that we should only be verifying unverified backups in the desired date range.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
+++ b/priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
@@ -137,8 +137,7 @@ public class BackupVerification {
                 backupVerificationResult.remotePath =
                         snapshotLocation.subpath(1, snapshotLocation.getNameCount()).toString();
                 result.add(backupVerificationResult);
-            }
-            else {
+            } else {
                 BackupVerificationResult backupVerificationResult =
                         verifyBackup(metaProxy, backupMetadata);
                 if (logger.isDebugEnabled())
--- a/priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
+++ b/priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
@@ -109,7 +109,7 @@ public class BackupVerification {
     }
 
     public List<BackupVerificationResult> verifyAllBackups(
-            BackupVersion backupVersion, boolean force, DateRange dateRange)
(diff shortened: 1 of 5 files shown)


priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -108,6 +108,43 @@ private IMetaProxy getMetaProxy(BackupVersion backupVersion) {
         return Optional.empty();
     }
 
+    public List<BackupVerificationResult> verifyAllBackups(
+            BackupVersion backupVersion, boolean force, DateRange dateRange)
```
arunagrawal84: Where are we using this flag of `force` here? I thought force means we re-verify the backups. If it is not used here better to remove this flag.

priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
```diff
@@ -88,22 +89,32 @@ public void execute() throws Exception {
...
             // send notification that backup is uploaded.
-            logger.info(
-                    "Sending {} message for backup: {}",
-                    AbstractBackupPath.BackupFileType.SNAPSHOT_VERIFIED,
-                    verificationResult.get().snapshotInstant);
-            backupNotificationMgr.notify(verificationResult.get());
+            verificationResults
+                    .stream()
+                    .forEach(
+                            r -> {
```
arunagrawal84: nit: please pick more descriptive variable rather than `r`

arunagrawal84: nit: please pick more descriptive variable like `result` rather than `r`

arunagrawal84 (commented):
LGTM: Fix n shipit.

priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
schakrovorthy: Sure I'll remove it.

priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
schakrovorthy: Sure I will modify it.

PR #868, opened 2020-05-16
CASS-1799 Fixing the Priam config endpoints after they were broken du…

schakrovorthy:
…e to a dependency update.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
+++ b/priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
@@ -18,6 +18,7 @@ package com.netflix.priam.resources;
 
 import com.google.inject.Inject;
 import com.netflix.priam.PriamServer;
+import com.netflix.priam.utils.GsonJsonSerializer;
 import java.util.HashMap;
 import java.util.Map;
 import javax.ws.rs.GET;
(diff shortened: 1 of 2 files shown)


priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
```diff
@@ -51,14 +52,17 @@ private Response doGetPriamConfig(String group, String name) {
...
                 result.put(name, value.get(name));
-                return Response.ok(result, MediaType.APPLICATION_JSON).build();
+                return Response.ok(GsonJsonSerializer.getGson().toJson(result)).build();
             } else if (name != null) {
                 result.put("message", String.format("No such structured config: [%s]", name));
                 logger.error(String.format("No such structured config: [%s]", name));
-                return Response.status(404).entity(result).type(MediaType.APPLICATION_JSON).build();
+                return Response.status(404)
+                        .entity(GsonJsonSerializer.getGson().toJson(result))
+                        .type(MediaType.TEXT_PLAIN)
```
arunagrawal84: Why this has to be plain text even when we convert to JSON?

priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
```diff
@@ -89,11 +92,14 @@ public Response getProperty(
...
                 result.put(name, value);
-                return Response.ok(result, MediaType.APPLICATION_JSON).build();
+                return Response.ok(GsonJsonSerializer.getGson().toJson(result)).build();
             } else {
                 result.put("message", String.format("No such property: [%s]", name));
                 logger.error(String.format("No such property: [%s]", name));
-                return Response.status(404).entity(result).type(MediaType.APPLICATION_JSON).build();
+                return Response.status(404)
+                        .entity(GsonJsonSerializer.getGson().toJson(result))
+                        .type(MediaType.TEXT_PLAIN)
```
arunagrawal84: Same as other.

priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
schakrovorthy: Fair point I have modified it to APPLICATION_JSON now.

priam/src/main/java/com/netflix/priam/resources/PriamConfig.java
schakrovorthy: Fair point I have modified it to APPLICATION_JSON now.

PR #882, opened 2020-05-20
Preliminary refactoring in advance of backporting PropertiesFileTuner…

mattl-netflix:
… to 3.11.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -16,6 +16,7 @@
  */
 package com.netflix.priam.config;
 
+import com.google.common.collect.ImmutableSet;
 import com.google.inject.ImplementedBy;
 import com.netflix.priam.tuner.JVMOption;
 import com.netflix.priam.config.PriamConfiguration;
@@ -44,16 +45,6 @@ public interface IConfiguration {
      * @return Path to the home dir of Cassandra
      */
     String getCassHome();
-
-    /**
-     * Returns the path to the Cassandra configuration directory, this varies by distribution but the default
-     * of finding the yaml file and going up one directory is a reasonable default. If you have a custom Yaml
-     * Location that doesn't have all the other configuration files you may want to tune this.
-     * @return
-     */
-    default String getCassConfigurationDirectory() {
-        return new File(getYamlLocation()).getParentFile().getPath();
-    }
 
     String getYamlLocation();
 
@@ -805,11 +796,10 @@ public interface IConfiguration {
     }
 
     /**
-     * Return a comma delimited list of property files that should be tuned in the configuration
-     * directory by Priam. These files live relative to the configuration directory.
-     * @return A comma delimited list of relative file paths to the configuration directory
+     * Return a list of property file paths from the configuration directory by Priam that should be tuned.
+     * @return the files paths
      */
-    default String getTunablePropertyFiles() { return ""; }
+    default ImmutableSet<String> getTunablePropertyFiles() { return ImmutableSet.of(); }
 
     /**
      * Escape hatch for getting any arbitrary property by key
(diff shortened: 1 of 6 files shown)


priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -809,7 +800,7 @@ default int getPostRestoreHookHeartbeatCheckFrequencyInMs() {
      * directory by Priam. These files live relative to the configuration directory.
      * @return A comma delimited list of relative file paths to the configuration directory
```
sumanth-pasupuleti: nit: update the comment for return type.

sumanth-pasupuleti:
@mattl-netflix  I see the pull request is against 4.x branch, vs 3.11?

mattl-netflix:
> @mattl-netflix I see the pull request is against 4.x branch, vs 3.11?

This refactoring can be applied to both so I thought I'd do it in 4.x and then cherry-pick to 3.11. Since it is not client facing, it can be just silently included in the next 4.x release whenever that comes.

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
mattl-netflix: Done

priam/src/main/java/com/netflix/priam/tuner/PropertiesFileTuner.java
```diff
@@ -1,87 +1,55 @@
...
-            properties.getLayout().load(properties, new FileReader(propertiesFile.getPath()));
-
-            Set<String> keys = new HashSet<>();
-            properties.getKeys().forEachRemaining(keys::add);
-
-            Map<String, String> overridenProperties = new HashMap<>();
-            String propertyOverrides = config.getProperty("propertyOverrides." + propertyPrefix, null);
-
-            if (propertyOverrides != null) {
+            properties.getLayout().load(properties, new FileReader(propertyFile));
```
schakrovorthy, line 40: What happens if the propertyFile does not exist? Is that handled by the FileReader in some way.

schakrovorthy: Never mind it seems it is being handled in the exception handler below.

priam/src/main/java/com/netflix/priam/tuner/PropertiesFileTuner.java
```diff
@@ -1,87 +1,55 @@
...
+            String overrides = config.getProperty("propertyOverrides." + FilenameUtils.getBaseName(propertyFile), null);
+            if (overrides != null) {
```
schakrovorthy, line 42: Would this code handle overrides being an empty string?

mattl-netflix: Good catch. It throws an IllegalArgumentException which I have confirmed in practice. Moreover, it will throw when overrides resolves to a map with missing or duplicate keys or missing values as well. This will cause Priam to fail to tune the yaml and report unhealthy. 

I assert this is the correct behavior as Priam will continue to start, thereby minimizing the potential impact on other scheduled jobs like backups, but also doesn't just fail silently as was the case previously.

schakrovorthy: That makes sense, is it possible to implement a default behavior if overrides is an empty string?

mattl-netflix: What default behavior do you think is best? 

My preference would be to ensure the following hold:

1) different user inputs result in different behavior
2) overrides overwrite specific properties in the file as specified in the String 

Assuming we adhere to these precepts, I think throwing on empty string is best because it prevents users from mistakenly thinking they've overwritten something when they haven't.

schakrovorthy:
Makes sense, looks good to me.

PR #887, opened 2020-06-19
CASS-1752 First pass of changes to throw an exception when a node is …

schakrovorthy:
…bootstrapping to an existing token. Per current thoughts this change is going to held as a PR and only committed after the root cause is addressed in Cassandra if any.


mattl-netflix (approved):

mattl-netflix (commented):
Reminder that this needs to be tested locally. Does this cause Priam to restart or report unhealthy or something else?

PR #889, opened 2020-06-24
CASS-1752 First pass of changes to throw an exception when a node is …

schakrovorthy:
…bootstrapping to an existing token. Per current thoughts this change is going to held as a PR and only committed after the root cause is addressed in Cassandra if any.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
+++ b/priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
@@ -35,7 +35,7 @@ public class TokenRetrieverUtils {
      */
     public static String inferTokenOwnerFromGossip(
             List<? extends PriamInstance> allIds, String token, String dc)
-            throws GossipParseException {
+            throws GossipParseException, TokenAliveException {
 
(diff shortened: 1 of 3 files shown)


priam/src/main/java/com/netflix/priam/identity/token/TokenRetrieverUtils.java
```diff
@@ -155,4 +156,22 @@ public GossipParseException(String message, Throwable t) {
             super(message, t);
         }
     }
+
+    /** This exception is thrown either when a node is bootstrapping using a token that is alive. */
+    public static class TokenAliveException extends Exception {
+
+        private static final long serialVersionUID = 1038678311186020257L;
+
+        public TokenAliveException() {
```
mattl-netflix, line 165: Looks like codecov is complaining because these overloads are not tested. They aren't called so I recommend to just remove them.

Specifically:

public TokenAliveException()
public TokenAliveException(String message, Throwable t)

mattl-netflix (approved):

mattl-netflix (changes requested):
Reminder that this needs to be tested locally. Does this cause Priam to restart or report unhealthy or something else?

PR #900, opened 2020-08-04
Throw when gossip unanimously says token is already owned by a live node

mattl-netflix:
Also:

Throw when there is a gossip mismatch.
Replace IP from database when not enough nodes respond.


arunagrawal84:
LGTM

arunagrawal84 (approved):

PR #906, opened 2020-09-21
CASS-1906 Backup secondary index files

mattl-netflix:


arunagrawal84:
@mattl-netflix Rejecting this PR with following issues- 
1. SnapshotMetaTask (new backup system) is not changed in this PR. 
2. Incremental path will upload index files at the wrong location (inside CF folder) due to this - https://github.com/Netflix/Priam/blob/3.11/priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java#L94
We only parse columnfamily and keyspace but never name of the index (change in C* 3.x from 2.x). 
This is where we need to start making changes where we create another entry for indexes. 
3. Restore path will fail due to above issue. 
4. Any backup will get corrupt as file names of column family could clash with index file names. We need to create that distinction everywhere.

PR #908, opened 2020-09-23
Stop filtering out OpsCenter keyspace

mattl-netflix:


mattl-netflix:
Merging this as it is a copy of the approved changes to 3.x.

PR #911, opened 2020-10-08
Backup Secondary Indexes

mattl-netflix:
Move CompressionAlgorithm and EncryptionAlgorithm to separate classes.
Remove dead code from AbstractBackupPath.
Style tweaks.
Add support for secondary index file type.
Remove redundant methods from FileUploadResult.
Remove redundant methods from ColumnFamilyResult
Inline PrefixGenerator's only method, unnest try-catch blocks, Use ImmutableSetMultimap where applicable.
Upload secondary index files for V2 backups


priam/src/main/java/com/netflix/priam/compress/ICompression.java
```diff
@@ -25,12 +25,6 @@
 @ImplementedBy(SnappyCompression.class)
 public interface ICompression {
 
-    enum CompressionAlgorithm {
```
sumanth-pasupuleti, line 28: What are the motivations behind moving CompressionAlgorithm and CryptographyAlgorithm to separate files from the interface files

sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/compress/ICompression.java
mattl-netflix: Immediately, this is only a readability optimization since you don't have to qualify the name of the enum when it is used.  Which makes sense. It's not like there are multiple CompressionAlogrithm enums that need to be disambiguated. 

More generally, a dependency on the CompressionAlgorithm enum is not the same thing as a dependency on the ICompression interface and so decoupling them eases future package refactoring. 

There's a short discussion here:  https://softwareengineering.stackexchange.com/questions/164884/best-practice-for-packing-java-enums

PR #915, opened 2021-01-15
Remove redundant information about bootstrapping from InstanceState

mattl-netflix:


sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/health/InstanceState.java
```diff
@@ -225,8 +180,9 @@ public void setYmlWritten(boolean yml) {
 
     public static class RestoreStatus {
         private LocalDateTime startDateRange, endDateRange; // Date range to restore from
-        private LocalDateTime executionStartTime,
-                executionEndTime; // Start-end time of the actual restore execution
+        // Start-end time of the actual restore execution
+        // Note these are referenced when this class is serialized to a String.
+        private LocalDateTime executionStartTime, executionEndTime;
```
arunagrawal84, line 185: At some point, we should remove LocalDateTime and replace with Instant.

mattl-netflix: Ack.

PR #922, opened 2021-03-12
Skip backup compression based on configuration

mattl-netflix:
Introduced new BackupsToCompress enum. ALL = compress all backups, NONE = compress no backups, UNCOMPRESSED = compress only those sstables lacking a CompressionInfo.db component.

In the process made AbstractBackupPath more of a first-class citizen and changed the IBackupFileSystem API's downloadFile method take one directly rather than Java Paths.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
@@ -29,7 +29,10 @@ import com.netflix.priam.utils.DateUtil;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
-import java.util.*;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
 
 /**
  * Represents location of an object on the remote file system. All the objects will be keyed with a
--- a/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
@@ -16,6 +16,7 @@
  */
 package com.netflix.priam.aws;
 
+import com.google.api.client.util.Lists;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -28,9 +29,7 @@ import com.netflix.priam.utils.DateUtil;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.Optional;
+import java.util.*;
 
 /**
  * Represents location of an object on the remote file system. All the objects will be keyed with a
@@ -103,16 +102,23 @@ public class RemoteBackupPath extends AbstractBackupPath {
         type = BackupFileType.valueOf(remotePath.getName(index++).toString());
         String lastModified = remotePath.getName(index++).toString();
         setLastModified(Instant.ofEpochMilli(Long.parseLong(lastModified)));
+        List<String> parts = Lists.newArrayListWithCapacity(4);
         if (BackupFileType.isDataFile(type)) {
             keyspace = remotePath.getName(index++).toString();
             columnFamily = remotePath.getName(index++).toString();
+            parts.add(keyspace);
+            parts.add(columnFamily);
         }
         if (type == BackupFileType.SECONDARY_INDEX_V2) {
             indexDir = remotePath.getName(index++).toString();
+            parts.add(indexDir);
         }
         setCompression(CompressionAlgorithm.valueOf(remotePath.getName(index++).toString()));
         setEncryption(remotePath.getName(index++).toString());
         fileName = remotePath.getName(index).toString();
+        parts.add(fileName);
+        this.backupFile =
+                Paths.get(config.getDataFileLocation(), parts.toArray(new String[] {})).toFile();
     }
 
     private String getV1Location() {
--- a/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import com.netflix.priam.backup.AbstractBackupPath;
+import com.netflix.priam.compress.CompressionAlgorithm;
 import com.netflix.priam.config.IConfiguration;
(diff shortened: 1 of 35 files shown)


priam/src/test/java/com/netflix/priam/backup/TestAbstractBackup.java
```diff
@@ -120,7 +120,8 @@ public String getName() {
     public void testCorrectCompressionAlgorithm() throws Exception {
         File parent = new File(DIRECTORY);
         AbstractBackupPath.BackupFileType backupFileType = AbstractBackupPath.BackupFileType.SST_V2;
-        List<AbstractBackupPath> abps = abstractBackup.upload(parent, backupFileType, false, false);
+        ImmutableSet<AbstractBackupPath> abps =
```
sumanth-pasupuleti: nit: this is pre-existing, but good to avoid abbreviated variable names.

priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
```diff
@@ -28,9 +29,7 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.Optional;
+import java.util.*;
```
sumanth-pasupuleti: curious if this change to .* was intentional

priam/src/main/java/com/netflix/priam/backupv2/FileUploadResult.java
```diff
@@ -47,6 +51,19 @@ public FileUploadResult(
...
+        this.compression = CompressionAlgorithm.SNAPPY;
+        this.encryption = CryptographyAlgorithm.PLAINTEXT;
+    }
+
+    public FileUploadResult(AbstractBackupPath path) {
+        File file = path.getBackupFile();
+        this.fileName = file.toPath();
+        this.backupPath = path.getRemotePath();
+        this.lastModifiedTime = path.getLastModified();
+        this.fileCreationTime = path.getLastModified(); // Remove. Always immutable in practice.
```
sumanth-pasupuleti: I agree with the comment that the file is immutable in practice. Curious on what you indicated to remove

priam/src/main/java/com/netflix/priam/backup/IBackupFileSystem.java
```diff
@@ -29,21 +29,20 @@
...
-     * @param remotePath fully qualified location of the file on remote file system.
-     * @param localPath location on the local file sytem where remote file should be downloaded.
+     * @param path Backup path representing a local and remote file pair
      * @param retry No. of times to retry to download a file from remote file system. If &lt;1, it
      *     will try to download file exactly once.
      * @throws BackupRestoreException if file is not available, downloadable or any other error from
      *     remote file system.
      */
-    void downloadFile(Path remotePath, Path localPath, int retry) throws BackupRestoreException;
+    void downloadFile(AbstractBackupPath path, String suffix, int retry)
```
sumanth-pasupuleti, line 38: It maybe helpful to call out this interface change explicitly in the release notes.

sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
```diff
@@ -80,48 +74,62 @@ private AbstractBackupPath getAbstractBackupPath(final File file, final BackupFi
...
+    }
+
+    private CompressionAlgorithm getCompressionAlgorithm(String file, Set<String> compressedFiles) {
+        BackupsToCompress which = config.getBackupsToCompress();
+        switch (which) {
+            case NONE:
+                return CompressionAlgorithm.NONE;
+            case ALL:
+                return CompressionAlgorithm.SNAPPY;
+            case UNCOMPRESSED:
```
arunagrawal84: nit: Can we name it more like IF_REQUIRED. Uncompressed does not show true intent.

arunagrawal84: Also thinking aloud, they are not compressionAlgorithms (like snappy/lz4) but compressionType.

arunagrawal84 (approved):

priam/src/main/java/com/netflix/priam/backup/IBackupFileSystem.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/aws/RemoteBackupPath.java
mattl-netflix: It was not. Fixed in future commit.

priam/src/main/java/com/netflix/priam/backupv2/FileUploadResult.java
mattl-netflix: Clarified thusly:

// TODO Remove this. It is redundant with fileCreationTime as the files are immutable.

priam/src/test/java/com/netflix/priam/backup/TestAbstractBackup.java
mattl-netflix: Fixed.

priam/src/main/java/com/netflix/priam/backup/AbstractBackup.java
mattl-netflix: I oppose this actually. IF_REQUIRED does not say which backups should be compressed. Is it a function of the file's size? NIC throughput? Number of cores?

I initially had it as:

Priam.whenToCompressBackups
enum WhenToCompress {
  ALWAYS
  NEVER
  WHEN_UNCOMPRESSED
}

but it led to fatter code for what I considered to be a minimal readability improvement. Let me know if you prefer that or insist upon IF_REQUIRED.

mattl-netflix: In your second comment, Are you saying the CompressionAlgorithm enum should be renamed?

arunagrawal84: I prefer elaborate rather than cryptic "uncompressed" which could mean to the operator that it is uncompressed rather than choosing NONE. And yes, I prefer renaming the enum as compression algo generally refer to SNAPPY, LZ4, xxHash etc.

priam/src/main/java/com/netflix/priam/backupv2/FileUploadResult.java
arunagrawal84: @mattl-netflix can you confirm if that statement is indeed true for 3.x (considering this PR is for 3.x). I know for fact that is it NOT true for C* 2.x as I have seen files that were created hrs/days before it was last written to. We actually had an incident due to this ^^

PR #936, opened 2021-04-21
Verify thrift listening on port before returning its status

ayushisingh29:
Priam marks isThriftActive = true if nodestool statusthrift is true. It does not check whether a process is listening on the rpc port. As a result Priam reports healthy even when nothing is actually listening on the rpc_port. 

With the changes added to this PR, Priam will only report isthriftActive = true if nodestool statusthrift is true and if something is listening on the rpc_port. This is added as a fast property and can be turned off by setting ".checkThriftOnPort" as false.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -1100,8 +1100,8 @@ public interface IConfiguration {
     }
 
     /** @return true to check is thrift listening on the rpc_port. */
-    default boolean isCheckThriftOnPortEnabled() {
-        return true;
+    default boolean checkThriftServerIsListening() {
+        return false;
     }
 
     /**
--- a/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/IConfiguration.java
@@ -1099,6 +1099,11 @@ public interface IConfiguration {
         return getSnitch().equals("org.apache.cassandra.locator.GossipingPropertyFileSnitch");
     }
 
+    /** @return true to check is thrift listening on the rpc_port. */
+    default boolean isCheckThriftOnPortEnabled() {
+        return true;
+    }
+
     /**
      * Escape hatch for getting any arbitrary property by key This is useful so we don't have to
      * keep adding methods to this interface for every single configuration option ever. Also
--- a/priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
@@ -765,7 +765,7 @@ public class PriamConfiguration implements IConfiguration {
     }
 
     @Override
-    public boolean isCheckThriftOnPortEnabled() {
-        return config.get(PRIAM_PRE + ".checkThriftOnPort", true);
+    public boolean checkThriftServerIsListening() {
+        return config.get(PRIAM_PRE + ".checkThriftServerIsListening", false);
     }
 }
--- a/priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
+++ b/priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
@@ -763,4 +763,9 @@ public class PriamConfiguration implements IConfiguration {
     public boolean isForgottenFileMoveEnabled() {
         return config.get(PRIAM_PRE + ".forgottenFileMoveEnabled", false);
     }
+
+    @Override
+    public boolean isCheckThriftOnPortEnabled() {
+        return config.get(PRIAM_PRE + ".checkThriftOnPort", true);
+    }
 }
--- a/priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
+++ b/priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
@@ -92,7 +92,10 @@ public class CassandraMonitor extends Task {
                 NodeProbe bean = JMXNodeTool.instance(this.config);
                 instanceState.setIsGossipActive(bean.isGossipRunning());
                 instanceState.setIsNativeTransportActive(bean.isNativeTransportRunning());
-                instanceState.setIsThriftActive(bean.isThriftServerRunning());
+                instanceState.setIsThriftActive(
+                        bean.isThriftServerRunning()
+                                && (!config.isCheckThriftOnPortEnabled()
+                                        || isListeningOnPort(config.getThriftPort())));
             } else {
                 // Setting cassandra flag to false
                 instanceState.setCassandraProcessAlive(false);
@@ -171,5 +174,20 @@ public class CassandraMonitor extends Task {
     public static void setIsCassadraStarted() {
         // Setting cassandra flag to true
         isCassandraStarted.set(true);
+    }
+
+    private boolean isListeningOnPort(int port) {
+        String[] cmd = {"/bin/sh", "-c", "ss -tuln | grep -c " + port};
+        String line = null;
+        try {
+            Process p = Runtime.getRuntime().exec(cmd);
+            p.waitFor();
+            BufferedReader b = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            line = b.readLine();
+            b.close();
+        } catch (Exception e) {
+            logger.warn("Exception thrown while checking if process is listening on a port ", e);
+        }
+        return line != null && Integer.parseInt(line) != 0;
     }
 }
--- a/priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
+++ b/priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
@@ -25,10 +25,7 @@ import com.netflix.priam.merics.CassMonitorMetrics;
 import com.netflix.priam.scheduler.SimpleTimer;
 import com.netflix.priam.scheduler.Task;
 import com.netflix.priam.scheduler.TaskTimer;
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStreamReader;
+import java.io.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.cassandra.tools.NodeProbe;
 import org.apache.commons.io.IOUtils;
@@ -92,10 +89,10 @@ public class CassandraMonitor extends Task {
                 NodeProbe bean = JMXNodeTool.instance(this.config);
                 instanceState.setIsGossipActive(bean.isGossipRunning());
                 instanceState.setIsNativeTransportActive(bean.isNativeTransportRunning());
+                ThriftChecker thriftChecker = new ThriftChecker(this.config);
                 instanceState.setIsThriftActive(
-                        bean.isThriftServerRunning()
-                                && (!config.isCheckThriftOnPortEnabled()
-                                        || isListeningOnPort(config.getThriftPort())));
+                        bean.isThriftServerRunning() && thriftChecker.isThriftServerListening());
+
             } else {
                 // Setting cassandra flag to false
                 instanceState.setCassandraProcessAlive(false);
@@ -174,20 +171,5 @@ public class CassandraMonitor extends Task {
     public static void setIsCassadraStarted() {
         // Setting cassandra flag to true
         isCassandraStarted.set(true);
-    }
-
-    private boolean isListeningOnPort(int port) {
-        String[] cmd = {"/bin/sh", "-c", "ss -tuln | grep -c " + port};
-        String line = null;
-        try {
-            Process p = Runtime.getRuntime().exec(cmd);
-            p.waitFor();
-            BufferedReader b = new BufferedReader(new InputStreamReader(p.getInputStream()));
-            line = b.readLine();
-            b.close();
-        } catch (Exception e) {
-            logger.warn("Exception thrown while checking if process is listening on a port ", e);
-        }
-        return line != null && Integer.parseInt(line) != 0;
     }
 }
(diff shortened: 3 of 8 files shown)


priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -1099,6 +1099,11 @@ default boolean usePrivateIP() {
         return getSnitch().equals("org.apache.cassandra.locator.GossipingPropertyFileSnitch");
     }
 
+    /** @return true to check is thrift listening on the rpc_port. */
+    default boolean isCheckThriftOnPortEnabled() {
+        return true;
```
mattl-netflix: booleans are best false by default.

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -1099,6 +1099,11 @@ default boolean usePrivateIP() {
         return getSnitch().equals("org.apache.cassandra.locator.GossipingPropertyFileSnitch");
     }
 
+    /** @return true to check is thrift listening on the rpc_port. */
+    default boolean isCheckThriftOnPortEnabled() {
```
mattl-netflix: Prefer checkThriftServerIsListening here and everywhere.

priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
```diff
@@ -763,4 +763,9 @@ public int getForgottenFileGracePeriodDaysForRead() {
     public boolean isForgottenFileMoveEnabled() {
         return config.get(PRIAM_PRE + ".forgottenFileMoveEnabled", false);
     }
+
+    @Override
+    public boolean isCheckThriftOnPortEnabled() {
+        return config.get(PRIAM_PRE + ".checkThriftOnPort", true);
```
mattl-netflix: Make the policy false at first as well. We'll roll it out in stages via FP change starting in test.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -92,7 +92,10 @@ public void execute() throws Exception {
                 NodeProbe bean = JMXNodeTool.instance(this.config);
                 instanceState.setIsGossipActive(bean.isGossipRunning());
                 instanceState.setIsNativeTransportActive(bean.isNativeTransportRunning());
-                instanceState.setIsThriftActive(bean.isThriftServerRunning());
+                instanceState.setIsThriftActive(
+                        bean.isThriftServerRunning()
+                                && (!config.isCheckThriftOnPortEnabled()
```
mattl-netflix: thriftChecker.thriftServerIsListening()

This will make more sense with the comments below.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -172,4 +175,19 @@ public static void setIsCassadraStarted() {
         // Setting cassandra flag to true
         isCassandraStarted.set(true);
     }
+
+    private boolean isListeningOnPort(int port) {
```
mattl-netflix: As we are doing IO it is better for testing purposes to create an interface with the following signature:

public interface ThriftChecker {
  boolean thriftServerIsListening() throws IOException;
}

The implementation of which will inject a PriamConfiguration and check if it should do anything at all via checkThriftServerIsListening and what the rpc_port is via config.getThriftPort. Just copy the bulk of this function there leaving the try-catch here instead. This will facilitate mocking in tests.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -172,4 +175,19 @@ public static void setIsCassadraStarted() {
...
+        String[] cmd = {"/bin/sh", "-c", "ss -tuln | grep -c " + port};
+        String line = null;
+        try {
+            Process p = Runtime.getRuntime().exec(cmd);
```
mattl-netflix: While I'll look the other way if it enables you to pack a fluent interface into a single line, prefer full variable names in all other cases.

That means p --> process here and b --> reader below.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -172,4 +175,19 @@ public static void setIsCassadraStarted() {
...
+            p.waitFor();
```
mattl-netflix: p.waitFor() is a risk to hang forever. Prefer the version that specifies a timeout. 1 second is fine. Hard-coded is fine.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -172,4 +175,19 @@ public static void setIsCassadraStarted() {
         // Setting cassandra flag to true
         isCassandraStarted.set(true);
     }
+
+    private boolean isListeningOnPort(int port) {
+        String[] cmd = {"/bin/sh", "-c", "ss -tuln | grep -c " + port};
```
mattl-netflix: The waitFor call below got me thinking. In cases where stderr is not empty this would hang forever because we didn't consume it. Instead let's route it to /dev/null because we don't really care in this case.

I believe you can do that by preceding the 'ss' with 2>/dev/null (something like that). If there is a sleeker way to do it in Java then prefer that but it isn't worth more than a minute of exploration.

mattl-netflix: I would just inline these in the exec command too.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -172,4 +175,19 @@ public static void setIsCassadraStarted() {
...
+        try {
+            Process p = Runtime.getRuntime().exec(cmd);
+            p.waitFor();
+            BufferedReader b = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            line = b.readLine();
+            b.close();
+        } catch (Exception e) {
+            logger.warn("Exception thrown while checking if process is listening on a port ", e);
+        }
+        return line != null && Integer.parseInt(line) != 0;
```
mattl-netflix: While pessimism is the correct policy generally, our main risk in this patch is false positives creating unnecessary on-call noise. We're really just trying to catch the failure mode where thrift servers are oom-ing across the cluster so we can tell customers that we already know about it when they come to us.

To that end return true on any failure and leave a single line of comment explaining why.

mattl-netflix: Finally, returning false for nothing listening on the rpc_port is a peculiar failure mode if nodetool continues to report that the thrift server is running. I very rarely recommend this but I think we would benefit from a log line at INFO explaining why we are returning false.

Maybe:

logger.info("Could not find anything listening on the rpc port {}!", config.getRpcPort());

mattl-netflix (changes requested):

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -25,10 +25,7 @@
 import com.netflix.priam.scheduler.SimpleTimer;
 import com.netflix.priam.scheduler.Task;
 import com.netflix.priam.scheduler.TaskTimer;
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStreamReader;
+import java.io.*;
```
mattl-netflix: Refrain from doing this _unless_ this is being done automatically by goJF.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
```diff
@@ -92,10 +89,10 @@ public void execute() throws Exception {
                 NodeProbe bean = JMXNodeTool.instance(this.config);
                 instanceState.setIsGossipActive(bean.isGossipRunning());
                 instanceState.setIsNativeTransportActive(bean.isNativeTransportRunning());
+                ThriftChecker thriftChecker = new ThriftChecker(this.config);
```
mattl-netflix: Inject an IThriftChecker. Don't instantiate directly here.

priam/src/main/java/com/netflix/priam/health/ThriftChecker.java
```diff
@@ -0,0 +1,53 @@
...
+    public ThriftChecker(IConfiguration config) {
+        this.config = config;
+    }
+
+    public boolean isThriftServerListening() {
+        if (!config.checkThriftServerIsListening()) {
+            return true;
+        }
+        try {
+            Process process =
```
mattl-netflix: I take it back. This looks totally inelegant. Please return to the model you had before where you put the commands in their own array. Please put that "command array" just outside the try as well.

Lines 25-33 should account for a total of two lines.

priam/src/main/java/com/netflix/priam/health/ThriftChecker.java
```diff
@@ -0,0 +1,53 @@
...
+                                        "/bin/sh",
+                                        "-c",
+                                        "ss -tuln | grep -c " + config.getThriftPort(),
+                                        " 2>/dev/null"
+                                    });
+            process.waitFor(1, TimeUnit.SECONDS);
+            BufferedReader reader =
+                    new BufferedReader(new InputStreamReader(process.getInputStream()));
+            String line = reader.readLine();
+            reader.close();
```
mattl-netflix: Just thought of something. This is a leak waiting to happen. I _think_ the best thing to do here is something like this:

Process process;
try {
  process = ...
} catch () {
  // just log
}
and then use try-with-resources like so:
try (BufferedReader reader =
                    new BufferedReader(new InputStreamReader(process.getInputStream()));) {
  if (Integer.parseInt(reader.readLine()) == 0) {
    logger.info("Could not find anything listening on the rpc port {}!", config.getThriftPort());
     return false;
  }
} catch (Exception e) {
  // log with different message from above
}

Do this guarantees the readers all get closes in case of error in the readLine() call using modern Java constructs. It doesn't look great with the two try's, but that's the best I've got. Whatever you decide, I think the try-with-resources is a must.

priam/src/main/java/com/netflix/priam/health/ThriftChecker.java
```diff
@@ -0,0 +1,53 @@
...
+            if (Integer.parseInt(line) == 0) {
+                logger.info(
+                        "Could not find anything listening on the rpc port {}!",
+                        config.getThriftPort());
+                return false;
+            }
+        } catch (Exception e) {
+            logger.warn("Exception thrown while checking if process is listening on a port ", e);
+        }
+        // Returning true for exceptions as well because we do not want to generate unnecessary
```
mattl-netflix: No need to pack a pillow, they have them on the plane! Here's how I would word it:

// A quiet on-call is our top priority, err on the side of avoiding false positives by default.

priam/src/test/java/com/netflix/priam/health/TestThriftChecker.java
```diff
@@ -0,0 +1,93 @@
...
+import org.junit.Test;
+
+public class TestThriftChecker {
+    private IConfiguration config;
+
+    @Mocked private Process mockProcess;
+
+    @Before
+    public void TestThriftChecker() {
+        Injector injector = Guice.createInjector(new BRTestModule());
```
mattl-netflix: No need for the extra Guice dependency. Just instantiate config like this:

config = new FakeConfiguration();

priam/src/test/java/com/netflix/priam/health/TestThriftChecker.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import mockit.Expectations;
+import mockit.Mocked;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestThriftChecker {
+    private IConfiguration config;
```
mattl-netflix: Make FakeConfiguration instead of IConfiguration.

priam/src/test/java/com/netflix/priam/health/TestThriftChecker.java
```diff
@@ -0,0 +1,93 @@
...
+
```
mattl-netflix, line 16: Rather than creating a ThriftChecker in every test case, create a private variable here and set it in the @Before method.

priam/src/test/java/com/netflix/priam/health/TestThriftChecker.java
```diff
@@ -0,0 +1,93 @@
...
+        final InputStream mockOutput = new ByteArrayInputStream("1".getBytes());
+        new Expectations() {
+            {
+                mockProcess.getInputStream();
+                result = mockOutput;
+            }
+        };
+        // Mock out the ps call
+        final Runtime r = Runtime.getRuntime();
+        String[] cmd = {
```
mattl-netflix, line 49: Finally, you could probably simplify the test and make it a _little_ less brittle by removing this and making the expectations below work for any string.

priam/src/main/java/com/netflix/priam/health/CassandraMonitor.java
ayushisingh29: This must have been done automatically by goJF.

priam/src/test/java/com/netflix/priam/health/TestThriftChecker.java
ayushisingh29: The actual function has this cmd and expects the test to have same command in or to override it with the mock process. So leaving it as it is.

mattl-netflix (approved):

PR #937, opened 2021-04-28
Delete empty file in advance of uploading

ayushisingh29:
During upload of SSTable component file, Priam will try to acquire tokens from its internal rate limiter. The number of tokens must be >0. An exception is thrown when the file is empty and it tries again and again to upload it. Thus blocking up the backup thread.  With these changes we will delete the empty file and proceed to the next file for uploading.


mattl-netflix:
Closing as obsolete

PR #940, opened 2021-05-20
Ingress rule configurability

mattl-netflix:
Make it possible to:
1. skip deleting ingress rules
2. skip updating ingress rules
3. only add ingress rules for public IPs

### DIFF
--- a/priam/src/main/java/com/netflix/priam/aws/AWSIPConverter.java
+++ b/priam/src/main/java/com/netflix/priam/aws/AWSIPConverter.java
@@ -1,0 +1,58 @@
+package com.netflix.priam.aws;
+
+import com.google.common.base.Splitter;
+import com.netflix.priam.identity.PriamInstance;
+import com.netflix.priam.identity.config.InstanceInfo;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AWSIPConverter implements IPConverter {
+    private static final Logger logger = LoggerFactory.getLogger(AWSIPConverter.class);
+    private static final Pattern IP_PART = Pattern.compile("^[0-9]{1,3}$");
+
+    private final InstanceInfo myInstance;
+
+    @Inject
+    public AWSIPConverter(InstanceInfo myInstance) {
+        this.myInstance = myInstance;
+    }
+
+    @Override
+    public Optional<String> getPublicIP(PriamInstance instance) {
+        if (instance.getInstanceId().equals(myInstance.getInstanceId())) {
+            return Optional.ofNullable(myInstance.getHostIP());
+        }
+        Optional<String> ip = extractIPFromHostnameString(instance.getHostName());
+        return !instance.getDC().equals(myInstance.getRegion()) && !ip.isPresent()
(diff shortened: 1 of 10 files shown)


priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -1111,6 +1111,26 @@ default BackupsToCompress getBackupsToCompress() {
...
+    /**
+     * @return true if Priam should skip updating ingress rules for ips found in the token database.
+     */
+    default boolean skipUpdatingOthersIngressRules() {
+        return false;
+    }
+
+    /** @return get the threshold at which point we might risk not getting our ingress rule set. */
+    default int getACLSizeWarnThreshold() {
+        return 500;
```
sumanth-pasupuleti: Should we have a slightly lower threshold like 400 or 450 (given 500 is the known aws limit)

priam/src/main/java/com/netflix/priam/aws/UpdateSecuritySettings.java
```diff
@@ -82,21 +77,25 @@ public void execute() {
...
-        if (!aclToRemove.isEmpty()) {
-            membership.removeACL(aclToRemove, port, port);
-            firstTimeUpdated = true;
+        if (!config.skipUpdatingOthersIngressRules()) {
+            Set<String> aclToAdd = Sets.difference(desiredAcl, currentAcl);
+            logger.info("ingress rules to update: {}", Joiner.on(",").join(aclToAdd));
+            if (!aclToAdd.isEmpty()) {
+                int resultingSize = aclToAdd.size() + currentAcl.size();
+                if (resultingSize > config.getACLSizeWarnThreshold()) {
+                    logger.warn("We are trying to make too many ingress rules! {}", resultingSize);
```
sumanth-pasupuleti: Should we also start emitting a metric for resultingSize (in addition to log line). Wondering how discoverable a log line alone would be.

priam/src/main/java/com/netflix/priam/aws/AWSIPConverter.java
```diff
@@ -0,0 +1,58 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AWSIPConverter implements IPConverter {
```
sumanth-pasupuleti, line 16: nit: good to add comments to public classes

sumanth-pasupuleti (approved):

priam/src/main/java/com/netflix/priam/aws/AWSIPConverter.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
mattl-netflix: I removed this altogether, began publishing warnings when we (possibly) fail to update ingress rules, and began publishing a gauge of the size of the list every time it is measured.

priam/src/main/java/com/netflix/priam/aws/UpdateSecuritySettings.java
mattl-netflix: Point taken. Done.

PR #944, opened 2021-05-20
Make disk_acess_mode tunable via fast property

ayushisingh29:
These changes will allow user to update Cassandra's disk_access_mode by changing the past property in Priam. Property associated with it is _app_name + "disk.access.mode"._


priam/src/main/java/com/netflix/priam/tuner/StandardTuner.java
```diff
@@ -128,6 +128,7 @@ public void writeAllProperties(String yamlLocation, String hostname, String seed
         map.put(
                 "compaction_large_partition_warning_threshold_mb",
                 config.getCompactionLargePartitionWarnThresholdInMB());
+        map.put("disk_access_mode", config.getDiskAccessMode());
```
mattl-netflix, line 131: If someone sets this to some gobbledygook do we know that Cassandra will fail to start?  If so, we can leave it as is, otherwise we should put in some validation. In a follow-up is fine.

ayushisingh29: For random values, cassandra will not start at all. It will fail with an exception:
```Caused by: org.yaml.snakeyaml.constructor.ConstructorException: null; Can't construct a java object for tag:yaml.org,2002:org.apache.cassandra.config.Config; exception=Cannot create property=disk_access_mode for JavaBean=org.apache.cassandra.config.Config@34bde49d; Unable to find enum value 'randomvalue' for enum class: org.apache.cassandra.config.Config$DiskAccessMode;  in 'reader', line 1, column 1:```

mattl-netflix:
I don't see a way to approve this but you have my blessing to resolve the conflicts and squash-and-merge.

PR #946, opened 2021-06-01
CASS-1608 derive public ip, toggle ingressing public ips exclusively

mattl-netflix:
Already proven on the 3.11 branch so I am merging without review.


mattl-netflix:
Merging without review as this has been confirmed working on 3.11 branch.

PR #947, opened 2021-06-01
Optionally skip compression when uploading backups

mattl-netflix:
This is proven working on 3.11 so merging without review.


mattl-netflix:
Merging as this has been confirmed working properly on 3.11 branch.

PR #948, opened 2021-06-08
Revert behavior back to truncating milliseconds in backups last modified time.

mattl-netflix:


mattl-netflix:
Happy to make changes pursuant to review comments. Committing now just to get an AMI baked overnight.

PR #955, opened 2021-07-03
Only upload secondary index directories if they exist in the main data directory.

mattl-netflix:
Only upload secondary index directories if they exist in the main data directory.


mattl-netflix:
This has been confirmed working. Releasing to be ready to push Monday morning. Will, of course, adjust then if there are any comments.

PR #961, opened 2021-08-04
CASS-2201 Ensure SI backup directories are deleted when empty

mattl-netflix:
Futures.whenAllComplete does not fire for collections of already complete futures. Correcting logic to accommodate this.


arunagrawal84 (approved):
LGTM.

PR #962, opened 2021-08-04
CASS-1752 Throw in case of gossip mismatch when determining replace_ip in assigned token case.

mattl-netflix:


arunagrawal84 (approved):
Please verify a case with only 3 nodes in a region having a mismatch (between the 2 nodes)

PR #975, opened 2021-11-30
Dynamic Rate Limiting of Snapshots

mattl-netflix:
Create a rate limiter that dynamically adjusts its throttle based on the bytes still to upload in all remaining snapshots and a user-specified target time. 

Adjust the throttle only when we've deviated by a user-configurable threshold to ensure the rate limiter is not constantly adjusted. In that case, it would be redundant as it does not throttle the subsequent file after an adjustment. 

Ensure that the target does not exceed the earlier of the next scheduled snapshot or the time at which we would fail to meet our backup verification SLO. 

One significant change of note is that we begin leveraging our call to AWS to determine whether a file exists when creating the meta file. Previously, we would perform a second redundant check when attempting to upload the file. Now we just delete files that are already there.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriterBuilder.java
+++ b/priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriterBuilder.java
@@ -65,7 +65,7 @@ public class MetaFileWriterBuilder {
     }
 
     public interface DataStep {
-        DataStep addColumnfamilyResult(
+        ColumnfamilyResult addColumnfamilyResult(
                 String keyspace,
                 String columnFamily,
                 ImmutableMultimap<String, AbstractBackupPath> sstables)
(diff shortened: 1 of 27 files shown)


priam/src/main/java/com/netflix/priam/backupv2/MetaFileWriterBuilder.java
```diff
@@ -65,7 +65,7 @@ public StartStep newBuilder() throws IOException {
     }
 
     public interface DataStep {
-        DataStep addColumnfamilyResult(
+        ColumnfamilyResult addColumnfamilyResult(
```
jrwest, line 68: nit: would rename to `ColumnFamilyResult` (capital F) to keep with Cassandra naming

jrwest (approved):
LGTM. One very minor nit.

PR #1002, opened 2022-09-12
Spread deletes over the interval depending on where the node is in the ring.

mattl-netflix:
Today, a node attempts to delete backups every backupTTLMonitorPeriod seconds from when Priam was started. This can lead to bunching across the fleet or a long time between attempts if Priam is restarted often enough. This PR makes the backupTTLTask take into account a node's position within the ring and use that to schedule its backup deletion as far away from other nodes. It also adds a random wait to ensure that if you're running a fleet with many clusters with a common topology, nodes in the same relative position don't fire all at the same time.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backupv2/BackupTTLTask.java
+++ b/priam/src/main/java/com/netflix/priam/backupv2/BackupTTLTask.java
@@ -20,14 +20,11 @@ package com.netflix.priam.backupv2;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.Singleton;
-import com.netflix.priam.backup.AbstractBackupPath;
-import com.netflix.priam.backup.BackupRestoreException;
-import com.netflix.priam.backup.IBackupFileSystem;
-import com.netflix.priam.backup.IFileSystemContext;
-import com.netflix.priam.backup.Status;
+import com.netflix.priam.backup.*;
 import com.netflix.priam.config.IBackupRestoreConfig;
 import com.netflix.priam.config.IConfiguration;
 import com.netflix.priam.health.InstanceState;
+import com.netflix.priam.identity.token.TokenRetriever;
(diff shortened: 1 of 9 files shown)


chengw-netflix (approved):

priam/src/main/java/com/netflix/priam/backupv2/BackupTTLTask.java
```diff
@@ -77,13 +76,18 @@ public BackupTTLTask(
...
+            throws Exception {
         super(configuration);
         this.backupRestoreConfig = backupRestoreConfig;
         this.metaProxy = metaProxy;
         this.fileSystem = backupFileSystemCtx.getFileStrategy(configuration);
         this.abstractBackupPathProvider = abstractBackupPathProvider;
         this.instanceState = instanceState;
+        this.maxWaitSeconds =
+                backupRestoreConfig.getBackupTTLMonitorPeriodInSec()
+                        / tokenRetriever.getRingPosition().getDenominator();
```
jrwest, line 90: is the intention here to evenly spread `maxWaitSeconds` across every node or is it to make it dependent on the ring position? If its the former, do we need to call an O(n) function to get a constant size and if its the latter should we be using `getNumerator()` instead of `getDenominator()`

priam/src/main/java/com/netflix/priam/backupv2/BackupTTLTask.java
```diff
@@ -102,6 +106,9 @@ public void execute() throws Exception {
             throw new Exception(JOBNAME + " already running");
         }
 
+        // Sleep a random amount but not so long that it will spill into the next token's turn.
+        if (maxWaitSeconds > 0) Thread.sleep(new Random().nextInt(maxWaitSeconds));
```
jrwest, line 110: should there be a lower bound here so we don't sleep e.g. 1 seconds? or is that little amount ok?

PR #1006, opened 2022-09-16
Adding ability to add more keys in the message attributes for backup notification messages

sthadeshwar:


priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -89,7 +94,7 @@ public void notify(BackupVerificationResult backupVerificationResult) {
     }
 
     private Map<String, MessageAttributeValue> getMessageAttributes(
-            AbstractBackupPath.BackupFileType backupFileType) {
+            AbstractBackupPath.BackupFileType backupFileType, JSONObject additionalAttributes) {
```
mattl-netflix: We can leave this signature as simple as we found it considering the JSONObject you pass in will always have s3clustername and backuptype as keys. Please remove the BackupFileType input. and get the "always-on" s3clustername and backuptype out of the JSONObject you pass in.

mattl-netflix: A note on naming. The JSONObject contains all attributes we're going to send in the SNS message not just additional one. Prefer message instead.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -77,7 +81,8 @@ public void notify(BackupVerificationResult backupVerificationResult) {
             jsonObject.put("snapshotInstant", backupVerificationResult.snapshotInstant);
```
mattl-netflix, line 78: As long as you're here, let's pay down some debt. Please convert the SNAPSHOT_VERIFIED string literal in line 80 to AbstractBackupPath.BackupFileType.SNAPSHOT_VERIFIED.name().

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
                 new MessageAttributeValue()
                         .withDataType("String")
                         .withStringValue(backupFileType.name()));
+
+        String additionalAttrs =
+                this.backupRestoreConfig.getBackupNotificationAdditionalMessageAttrs();
+        if (!StringUtils.isBlank(additionalAttrs)) {
```
mattl-netflix: This will be made redundant. Remove it.

priam/src/main/java/com/netflix/priam/config/IBackupRestoreConfig.java
```diff
@@ -113,4 +113,15 @@ default boolean enableV2Restore() {
...
-}
+
+    /**
+     * Returns ',' separated list of attribute names. If not empty, adds the specified attributes to
+     * MessageAttributes in the backup notifications. SNS filter policy needs keys in
+     * MessageAttributes in order to filter based on those keys.
+     *
+     * @return
+     */
+    default String getBackupNotificationAdditionalMessageAttrs() {
```
mattl-netflix: Make this return a non-null ImmutableList(). The default should be an empty list.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+            try {
+                List<String> attrs =
```
mattl-netflix: Move this Splitting logic to BackupRestoreConfig. Be mindful to handle nulls properly.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+                        Splitter.on(",")
+                                .omitEmptyStrings()
+                                .trimResults()
+                                .splitToList(additionalAttrs);
+                for (String attr : attrs) {
+                    Object value = additionalAttributes.get(attr);
+                    if (value != null) {
+                        // Does not handle all data types
```
mattl-netflix: Remove the comment. It is plain to see from the code.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+                        if ("Instant".equals(value.getClass().getSimpleName())) {
```
mattl-netflix: Remove this block as it will never get hit. Note that Instant is not a valid type to put into a JSONObject. 

Per the Javadoc of JSONObject.put(String, Object):

* @param value An object which is the value. It should be of one of these
     *  types: Boolean, Double, Integer, JSONArray, JSONObject, Long, String,
     *  or the JSONObject.NULL object.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+        if (!StringUtils.isBlank(additionalAttrs)) {
+            try {
+                List<String> attrs =
+                        Splitter.on(",")
+                                .omitEmptyStrings()
+                                .trimResults()
+                                .splitToList(additionalAttrs);
+                for (String attr : attrs) {
+                    Object value = additionalAttributes.get(attr);
+                    if (value != null) {
```
mattl-netflix: Prefer message.has() for readability.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+        String additionalAttrs =
+                this.backupRestoreConfig.getBackupNotificationAdditionalMessageAttrs();
+        if (!StringUtils.isBlank(additionalAttrs)) {
+            try {
+                List<String> attrs =
+                        Splitter.on(",")
+                                .omitEmptyStrings()
+                                .trimResults()
+                                .splitToList(additionalAttrs);
+                for (String attr : attrs) {
```
mattl-netflix: Prefer modern Java unless you have performance reasons to do otherwise.  So that means:

backupRestoreConfig.getBackupNotificationAdditionalMessageAttrs()
.stream()
.filter(attr -> message.has(attr))
.forEach(attr -> messageAttributes.put(attr, toStringAttribute(message.get(attr));

(or the version of that that actually compiles)

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+                            byte[] byteValue = SerializationUtils.serialize((Instant) value);
+                            messageAttributes.putIfAbsent(
+                                    attr,
+                                    new MessageAttributeValue()
+                                            .withDataType("Binary")
+                                            .withBinaryValue(ByteBuffer.wrap(byteValue)));
+                        } else {
+                            messageAttributes.putIfAbsent(
+                                    attr,
+                                    new MessageAttributeValue()
```
mattl-netflix: There are now enough instances of this that it warrants its own function:

private MessageAttributeValue toStringAttribute(String value) {
    return new MessageAttributeValue().withDataType("String").withStringValue(value);
}

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -102,6 +107,42 @@ private Map<String, MessageAttributeValue> getMessageAttributes(
...
+                                            .withDataType("String")
+                                            .withStringValue(value.toString()));
+                        }
+                    }
+                }
+            } catch (Exception e) {
```
mattl-netflix: I'm not sure what you're expecting to catch here. Please remove the try-catch.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
```diff
@@ -89,7 +94,7 @@ public void notify(BackupVerificationResult backupVerificationResult) {
     }
 
     private Map<String, MessageAttributeValue> getMessageAttributes(
-            AbstractBackupPath.BackupFileType backupFileType) {
+            AbstractBackupPath.BackupFileType backupFileType, JSONObject additionalAttributes) {
         // SNS Attributes for filtering messages. Cluster name and backup file type.
         Map<String, MessageAttributeValue> messageAttributes = new HashMap<>();
         messageAttributes.putIfAbsent(
```
mattl-netflix: As referenced above these 10 lines resolve to:

messageAttributes.put("s3clustername", toStringAttribute(message.get("s3clustername")));
messageAttributes.put("backuptype", toStringAttribute(message.get("backuptype")));

mattl-netflix (changes requested):

priam/src/main/java/com/netflix/priam/config/IBackupRestoreConfig.java
sthadeshwar: I am following the convention of other config methods which also return a String. Are you sure you want me to change the return type as List? For that, I will also have to call getList() instead of get() on IConfigSource. I would suggest leave it as is and parse the String while reading (which is what getBackupNotifyComponentIncludeList() also does).

mattl-netflix: Returning a concatenated String to stand in as a list of Strings is not the best choice considering all callers are just going to split it anyway. If you insist on strictly adhering to convention, change it to a sensible one everywhere in a preliminary refactoring. I don't insist upon that though.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
sthadeshwar: I prefer idiomatic Java as well wherever I can but when there are checked exceptions thrown from the code being called, it does not make it look pretty. There is no performance benefit that I know of for using for loop instead of forEach in this case.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
sthadeshwar: I was just trying to be safe. Didn't want any bug or issue in parsing config to cause the notifications to be not sent.

sthadeshwar: Will remove it.

priam/src/main/java/com/netflix/priam/notification/BackupNotificationMgr.java
sthadeshwar: Done.

PR #1013, opened 2022-11-16
Create operator-specifiable time such that if a backup file was written before then it is automatically compressed using SNAPPY before upload.

mattl-netflix:


chengw-netflix (commented):
+1. LGTM

PR #1015, opened 2022-12-05
Log error message before notifying of failure

akhaku:
I suspect that the notify line is throwing since I see metrics indicating that an upload failed but no log message.


akhaku:
Separately, I wonder if we should have stuck with the defensive pattern here https://github.com/Netflix/Priam/pull/1006#discussion_r974560234 perhaps with a RuntimeException?
Or perhaps we should do something more holistic and handle any exceptions thrown in any EventObserver method in our `notify*` methods.

sthadeshwar (approved):

akhaku:
superseded by #1019

PR #1016, opened 2022-12-05
Log and swallow exceptions during backup notification

akhaku:
This is a follow-up to https://github.com/Netflix/Priam/pull/1015#issuecomment-1337967748


akhaku:
Superseded by #1019

PR #1017, opened 2022-12-06
Log backup failures, don't just ignore them

akhaku:


priam/src/main/java/com/netflix/priam/backupv2/SnapshotMetaTask.java
```diff
@@ -285,7 +287,26 @@ private void uploadAllFiles(final File backupDir) throws Exception {
                 // We do not want to wait for completion and we just want to add them to queue. This
                 // is to ensure that next run happens on time.
                 AbstractBackupPath.BackupFileType type = AbstractBackupPath.BackupFileType.SST_V2;
-                backupHelper.uploadAndDeleteAllFiles(snapshotDirectory, type, target, true);
+                backupHelper
+                        .uploadAndDeleteAllFiles(snapshotDirectory, type, target, true)
+                        .forEach(
+                                future ->
+                                        Futures.addCallback(
```
akhaku, line 294: not the biggest fan of this sparse formatting, but that's what `googleJavaFormat` did

priam/src/main/java/com/netflix/priam/backupv2/SnapshotMetaTask.java
```diff
@@ -285,7 +287,27 @@ private void uploadAllFiles(final File backupDir) throws Exception {
...
+                                                future,
+                                                new FutureCallback<AbstractBackupPath>() {
+                                                    @Override
+                                                    public void onSuccess(
+                                                            AbstractBackupPath result) {}
+
+                                                    @Override
+                                                    public void onFailure(Throwable t) {
+                                                        logger.error(
```
akhaku, line 303: I think we should be doing more than logging an error here, probably failing the backup entirely and not uploading a manifest, since the backup is corrupted.
This is just a start because right now we don't log the `BackupRestoreException` anywhere, we just propagate it up the stack and it disappears here.

mattl-netflix: All virgin mutations are accounted for in the incremental backups. For snapshot backups it is acceptable for one to fail from time to time as long as we don't go too long without a healthy one. Remember, the errant file would need to be a recent compaction product and then be compacted away before the next attempt. Given our low rate of failures this is an acceptable state to be in. I don't know what besides metrics and logging we should do TBH. Upload attempts are retried 10x in practice.

akhaku: We should at least have a log message when something fails, rather than fail silently. Perhaps we can drop it to WARN instead of ERROR?

akhaku: ah perhaps you were responding to my comment rather than saying we should change things here?

mattl-netflix: That's correct.

akhaku:
Superseded by #1025

PR #1019, opened 2023-01-09
Remove redundant interfaces and swap log and notification lines

mattl-netflix:
This PR is organized to be easily digestible if each commit is reviewed in series. We remove the redundant BackupEvent, EventObserver and EventGenerator interfaces. This leaves us in a state where if we want to add special handling to Exceptions other than JSONExceptions in BackupNotificationMgr it is just a one-line change.


priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -221,13 +221,13 @@ public Long retriableCall() throws Exception {
...
             } catch (Exception e) {
                 backupMetrics.incrementInvalidUploads();
-                backupNotificationMgr.notify(path, UploadStatus.FAILED);
                 logger.error(
                         "Error while uploading file: {} to location: {}. Exception: Msg: [{}], Trace: {}",
                         localPath,
                         remotePath,
                         e.getMessage(),
                         e.getStackTrace());
+                backupNotificationMgr.notify(path, UploadStatus.FAILED);
```
akhaku, line 230: 👍 obviates #1015
any thoughts on incorporating the change to remove `e.getStackTrace()` and instead passing `e` as the last parameter to get the stack trace?

PR #1025, opened 2023-01-17
Log backup failures rather than ignoring them.

mattl-netflix:
This Ammar's change from PR 1017 refactored for style only.


priam/src/main/java/com/netflix/priam/backupv2/SnapshotMetaTask.java
```diff
@@ -427,4 +431,18 @@ private static Optional<String> getPrefix(File file) {
     void setSnapshotName(String snapshotName) {
         this.snapshotName = snapshotName;
     }
+
+    private void addCallback(ListenableFuture<AbstractBackupPath> future, File snapshotDir) {
```
akhaku: minor, perhaps make this `static`?

akhaku (approved):

priam/src/main/java/com/netflix/priam/backupv2/SnapshotMetaTask.java
mattl-netflix: Good call. Done.

PR #1029, opened 2023-02-27
Switch from com.google.inject to JSR-330 javax.inject annotations for better compatibility

chengw-netflix:
Switch from com.google.inject to JSR-330 javax.inject annotations for better compatibility


arunagrawal84:
ship it

arunagrawal84 (approved):

PR #1034, opened 2023-03-27
Fix backup verification race condition causing missing notifications

mattl-netflix:


priam/src/main/java/com/netflix/priam/backup/BackupVerification.java
```diff
@@ -65,7 +68,7 @@ public IMetaProxy getMetaProxy(BackupVersion backupVersion) {
         return null;
     }
 
-    public Optional<BackupVerificationResult> verifyBackup(
+    public Optional<BackupVerificationResult> findLatestVerifiedBackup(
```
akhaku: this also verifies the backup right, in the verifyBackup call at line 88 (new). It doesn't _just_ find the latest verified backup?

priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
```diff
@@ -84,23 +84,24 @@ public void execute() throws Exception {
...
+                                    result.getSnapshotLocation());
+                            backupNotificationMgr.notify(
+                                    result.getSnapshotLocation(), result.getStart().toInstant());
+                        });
 
-        if (!backupVerification
-                .findLatestVerifiedBackup(
-                        BackupVersion.SNAPSHOT_META_SERVICE, false /* force */, dateRange)
-                .isPresent()) {
+        if (verifiedBackups.isEmpty()) {
```
akhaku, line 104: nice, this is the bit that fixes the actual bug 👍

akhaku (approved):

priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
mattl-netflix: 315 lines of making the change easy. 5 lines of making the easy change.

PR #1035, opened 2023-03-28
Reveal hook to allow operators to restore just to the most recent snapshot

mattl-netflix:


priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -54,18 +54,11 @@ public BackupRestoreUtil(String configIncludeFilter, String configExcludeFilter)
...
-                return Optional.of(meta);
-            }
-        }
-
-        return Optional.empty();
+        return metaProxy
+                .findMetaFiles(dateRange)
+                .stream()
+                .filter(meta -> metaProxy.isMetaFileValid(meta).valid)
+                .findFirst();
```
akhaku, line 58: beautiful

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -133,22 +129,12 @@ public final boolean isFiltered(String keyspace, String columnFamilyDir) {
             if (excludeFilter.containsKey(keyspace)
                     && (excludeFilter.get(keyspace).isEmpty()
                             || excludeFilter.get(keyspace).contains(columnFamilyName))) {
-                logger.debug(
```
akhaku, line 167: I suppose it's fine and you have a better understanding around whether or not we ever need those log messages but in general I'm not opposed to debug logs that I can enable to help debug an issue

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -133,22 +129,12 @@ public final boolean isFiltered(String keyspace, String columnFamilyDir) {
...
-                        "Skipping: keyspace: {}, CF: {} is part of exclude list.",
-                        keyspace,
-                        columnFamilyName);
                 return true;
             }
         if (includeFilter != null)
-            if (!(includeFilter.containsKey(keyspace)
+            return !(includeFilter.containsKey(keyspace)
```
akhaku, line 139: minor nit: could collapse this further, `return includeFilter != null && !includeFilter.containsKey(keyspace) && foo || bar`

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -17,6 +17,7 @@
 
 package com.netflix.priam.backup;
 
+import com.google.api.client.util.Lists;
```
akhaku: I'm generally not a fan of using (non-shaded) guava in a library since it causes so many compatibility issues. I've gone as far as created my own `Preconditions` class. This usage below looks like it might be able to be replaced with a vanilla `ArrayList` constructor.

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -58,14 +59,13 @@ public static Optional<AbstractBackupPath> getLatestValidMetaPath(
                 .findFirst();
     }
 
-    public static List<AbstractBackupPath> getAllFiles(
+    public static List<AbstractBackupPath> getMostRecentSnapshotPaths(
```
akhaku, line 61: renaming the method was a great idea, makes it more precise (now that it's different) and also makes it impossible to forget to update a callsite

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
```diff
@@ -245,9 +245,11 @@ public void restore(DateUtil.DateRange dateRange) throws Exception {
...
                     BackupRestoreUtil.getMostRecentSnapshotPaths(
                             latestValidMetaFile.get(), metaProxy, pathProvider);
-            allFiles.addAll(
-                    BackupRestoreUtil.getIncrementalPaths(
-                            latestValidMetaFile.get(), dateRange, metaProxy));
+            if (!config.skipIncrementalRestore()) {
+                allFiles.addAll(
+                        BackupRestoreUtil.getIncrementalPaths(
+                                latestValidMetaFile.get(), dateRange, metaProxy));
+            }
```
akhaku, line 252: I take it we don't have to update the callsite in BackupServletV2?

akhaku (approved):
Nice work, gotta say I love all of the cleanup you're doing; extraneous comments, vertical space, and declared exceptions are in particular some of my pet peeves

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
mattl-netflix: That is a good point. However, I wanted to leave code in methods unrelated to the goal of the PR unchanged. I accepted this because it was done automatically by Intellij.  If I change this I'd need to add the appropriate tests and I think that is beyond scope and best left to a follow-up.

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
mattl-netflix: I updated it there as well.

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
mattl-netflix: Removed.

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
akhaku: I'm confused - do we need to use `config.skipIncrementalRestore` there too? I didn't see that there in this PR, perhaps it isn't required?

PR #1048, opened 2023-04-20
Increment cross regional duplicate tokens to replicate the policy we have been applying manually.

mattl-netflix:
Throw when duplicate tokens are created in region because that would be an obvious error and we should not add two nodes in the same region so closely together.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
+++ b/priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
@@ -190,7 +190,12 @@ public class TokenRetriever implements ITokenRetriever {
     private boolean newTokenIsADuplicate(String newToken, ImmutableSet<PriamInstance> instances) {
         for (PriamInstance priamInstance : instances) {
             if (newToken.equals(priamInstance.getToken())) {
-                Preconditions.checkState(!myInstanceInfo.getRegion().equals(priamInstance.getDC()));
+                if (myInstanceInfo.getRegion().equals(priamInstance.getDC())) {
+                    throw new IllegalStateException(
+                            String.format(
+                                    "Trying to add token %s to %s but it already exists in %s",
+                                    newToken, myInstanceInfo.getRegion(), priamInstance.getDC()));
+                }
                 return true;
             }
         }
--- a/priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
+++ b/priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
@@ -155,28 +155,46 @@ public class TokenRetriever implements ITokenRetriever {
                 set(100, 100);
                 logger.info("Trying to generate a new token");
                 sleeper.sleep(new Random().nextInt(15000));
-                String myRegion = myInstanceInfo.getRegion();
-                // this offset ensures the nodes are spread far away from the other regions.
-                int regionOffset = tokenManager.regionOffset(myRegion);
-                String myRac = myInstanceInfo.getRac();
-                List<String> racs = config.getRacs();
-                int mySlot =
-                        factory.getAllIds(config.getAppName())
-                                .stream()
-                                .filter(i -> i.getRac().equals(myRac))
-                                .map(PriamInstance::getId)
-                                .max(Integer::compareTo)
-                                .map(id -> racs.size() + Math.max(id, regionOffset) - regionOffset)
-                                .orElseGet(
-                                        () -> {
-                                            Preconditions.checkState(racs.contains(myRac));
-                                            return racs.indexOf(myRac);
-                                        });
-                int instanceCount = membership.getRacCount() * membership.getRacMembershipSize();
-                String newToken = tokenManager.createToken(mySlot, instanceCount, myRegion);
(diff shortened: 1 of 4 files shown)


priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
```diff
@@ -155,30 +155,48 @@ public PriamInstance retriableCall() throws Exception {
...
+        while (newTokenIsADuplicate(newToken, allIds)) {
+            newToken = new BigInteger(newToken).add(BigInteger.ONE).toString();
+        }
+        return createToken(mySlot + regionOffset, newToken);
+    }
+
+    private boolean newTokenIsADuplicate(String newToken, ImmutableSet<PriamInstance> instances) {
+        for (PriamInstance priamInstance : instances) {
+            if (newToken.equals(priamInstance.getToken())) {
+                Preconditions.checkState(!myInstanceInfo.getRegion().equals(priamInstance.getDC()));
```
chengw-netflix: maybe add a WARN log here for better understanding the cause of the duplicate tokens?

chengw-netflix (approved):
LGTM.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
mattl-netflix: Done.

It will crash and log on failure regardless. But you bring up a great point:  the failure because two region Strings don't match is totally confusing without looking at the code. I have updated the code to print a clearer error message.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
```diff
@@ -155,30 +155,53 @@ public PriamInstance retriableCall() throws Exception {
...
+                        .max(Integer::compareTo)
+                        .map(id -> racs.size() + Math.max(id, regionOffset) - regionOffset)
+                        .orElseGet(
+                                () -> {
+                                    Preconditions.checkState(racs.contains(myRac));
+                                    return racs.indexOf(myRac);
+                                });
+        int instanceCount = membership.getRacCount() * membership.getRacMembershipSize();
+        String newToken = tokenManager.createToken(mySlot, instanceCount, myRegion);
+        while (newTokenIsADuplicate(newToken, allIds)) {
```
akhaku, line 184: Not strictly related to your PR but: do we need to worry about atomicity between the check here and the create below? E.g. if this is backed by Cassandra, do we need to use LWTs?

priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
```diff
@@ -155,30 +155,53 @@ public PriamInstance retriableCall() throws Exception {
...
+            newToken = new BigInteger(newToken).add(BigInteger.ONE).toString();
+        }
+        return createToken(mySlot + regionOffset, newToken);
+    }
+
+    private boolean newTokenIsADuplicate(String newToken, ImmutableSet<PriamInstance> instances) {
+        for (PriamInstance priamInstance : instances) {
+            if (newToken.equals(priamInstance.getToken())) {
+                if (myInstanceInfo.getRegion().equals(priamInstance.getDC())) {
```
akhaku, line 193: I'm a little unclear as to why we throw an `IllegalStateException` in this case vs the case where the DCs are different - IIUC, duplicates across DCs never happen because of the region hash offset? So it feels that _that's_ the case we should throw vs return false

Also, I'm having trouble tracing the entire code path given the multiple impls, but is it possible that `InstanceInfo#getRegion()` will return `us-east-1` for `us-east-1` while `PriamInstance#getDC` will return `us-east`?

priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
mattl-netflix: Internally we don't use LWTs, but Contention is accounted for by "acquiring a lock" in a separate table before writing. Note that contention is managed based on the token's relative position within the region rather than the token value itself. That prevents this change introducing the possibility of multiple tokens that are one apart in the same region.

priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
mattl-netflix: > IIUC, duplicates across DCs never happen because of the region hash offset?

Duplicates across DCs can happen if the region hash is off by one and the tokens between regions are off by one in the opposite direction. That scenario can result if the cluster has been doubled using Priam's doubling facilities. The purpose of this PR is to adapt to this scenario. That said, there should never be in-region duplicates and if there are we need to fail immediately. In that case we should not increment until there are no duplicates because that will lead to a situation where tokens are one apart within the same region. That would cause a data imbalance.

> is it possible that InstanceInfo#getRegion() will return us-east-1 for us-east-1 while PriamInstance#getDC will return us-east?

This is an interesting concern, but I don't believe it is a threat. The only place we ever convert away from AWS region names is in an internal tuning use case. I can go into greater depth about that this morning if you're interested.

PR #1049, opened 2023-04-24
Change the interface of PriamScheduler

chengw-netflix:
This pull request modifies the interface of the PriamScheduler class by introducing a more general PriamJobFactory that replaces the GuiceJobFactory. This update eliminates the need to expose the Guice module and enhances compatibility with the Spring Boot framework.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/scheduler/PriamScheduler.java
+++ b/priam/src/main/java/com/netflix/priam/scheduler/PriamScheduler.java
@@ -29,15 +29,15 @@ import org.slf4j.LoggerFactory;
 public class PriamScheduler {
     private static final Logger logger = LoggerFactory.getLogger(PriamScheduler.class);
     private final Scheduler scheduler;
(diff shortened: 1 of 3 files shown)


priam/src/main/java/com/netflix/priam/scheduler/PriamScheduler.java
```diff
@@ -29,15 +29,15 @@
 public class PriamScheduler {
     private static final Logger logger = LoggerFactory.getLogger(PriamScheduler.class);
     private final Scheduler scheduler;
-    private final GuiceJobFactory jobFactory;
+    private final PriamJobFactory jobProvider;
```
mattl-netflix, line 32: Let's keep terms consistent. That is, use jobFactory rather than jobProvider. I know we're moving away from Guice, but Provider has a specific meaning in that context.

mattl-netflix (approved):

PR #1082, opened 2024-03-04
Remove Backup Version V1

mattl-netflix:
The commits are meant to be reviewed in order to keep the process tractable.

Remove unused CLI tools.
Remove tooling to back up commit logs. It is no longer supported.
Remove IFileSystemContext. It is redundant.
Remove BackupService
Remove BackupVersion.
Remove GCS-specific code. It is no longer in use.
Remove remaining vestiges of backup v1.


priam/src/test/java/com/netflix/priam/backup/TestBackupVerification.java
```diff
@@ -125,82 +109,23 @@ private void setUp() throws Exception {
...
-                                new DateRange(backupDate + "," + backupDate))
-                        .stream()
-                        .findFirst();
-        Assert.assertTrue(backupMetadata.isPresent());
-        Assert.assertNull(backupMetadata.get().getLastValidated());
+        backupStatusMgr.finish(getBackupMetaData(start, Status.FINISHED));
     }
 
     @Test
     public void verifyBackupVersion1DateRange() throws Exception {
```
ayushisingh29, line 116: It will be good if we can update the function name(remove version1) or remove this test completely as v2 test is already present in verifyBackupVersion2DateRange

priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
```diff
@@ -212,11 +212,6 @@ public String getMaxDirectMemory() {
                 (PRIAM_PRE + ".direct.memory.size.") + instanceInfo.getInstanceType(), "50G");
     }
 
-    @Override
-    public String getBackupCronExpression() {
```
ayushisingh29, line 216: Why is this removed?

chengw-netflix (approved):

priam/src/main/java/com/netflix/priam/config/PriamConfiguration.java
mattl-netflix: It was only referenced in SnapshotBackup which was removed in this cleanup. The modern equivalent is getSnapshotMetaServiceCronExpression() used in SnapshotMetaTask.

priam/src/test/java/com/netflix/priam/backup/TestBackupVerification.java
mattl-netflix: Agreed. The test is redundant and should be removed. I will do that in a follow-up.

PR #1094, opened 2024-05-05
Allow for in-place restores.

mattl-netflix:
6369b6aa 2024-05-02 | Cease stopping Cassandra in the restore process and reveal the ability to restore to a staging area and then import or move the data depending on whether Cassandra is running.
4ece8d3a 2024-05-01 | Move getBackupDirectories to BackupRestoreUtil and change its signature to cease requiring an IConfiguration but rather just take the data directory path as a String.
6b3d45a8 2024-04-10 | Cease wiping the data directory in advance of restore. It is counter-intuitive and will be empty in practice in most cases anyway.
f10289fd 2024-04-10 | Consolidate restore status API.
4d6d07f3 2024-03-09 | Remove backup racs as that configuration is not used in practice and is not relevant to whether a restore should be attempted.
aa3dffe7 2024-03-09 | Remove redundant vertical whitespace.
8c23800c 2024-03-09 | Remove redundant logging.
646973d8 2024-03-09 | Remove all non-javadoc comments that don't provide warnings to future maintainers.
8d18f511 2024-03-09 | Remove waitForCompletion parameter to download method which is always false in practice. Remove final modifier on static method. Make static property an instance variable instead. Use parameterized constructor call in place of two lines.
ce6c7c10 2024-03-09 | Remove unused code from our restore tooling.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
@@ -137,24 +137,24 @@ public abstract class AbstractBackupPath implements Comparable<AbstractBackupPat
 
     /** Local restore file */
     public File newRestoreFile() {
-        File return_;
+        File file;
         String dataDir = config.getRestoreDataLocation();
         switch (type) {
             case SECONDARY_INDEX_V2:
-                return_ =
+                file =
                         new File(
                                 PATH_JOINER.join(
                                         dataDir, keyspace, columnFamily, indexDir, fileName));
                 break;
             case META_V2:
-                return_ = new File(PATH_JOINER.join(config.getDataFileLocation(), fileName));
+                file = new File(PATH_JOINER.join(config.getDataFileLocation(), fileName));
                 break;
             default:
-                return_ = new File(PATH_JOINER.join(dataDir, keyspace, columnFamily, fileName));
+                file = new File(PATH_JOINER.join(dataDir, keyspace, columnFamily, fileName));
         }
-        File parent = new File(return_.getParent());
+        File parent = new File(file.getParent());
         if (!parent.exists()) parent.mkdirs();
-        return return_;
+        return file;
     }
 
     @Override
--- a/priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
+++ b/priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
@@ -138,12 +138,13 @@ public abstract class AbstractBackupPath implements Comparable<AbstractBackupPat
     /** Local restore file */
     public File newRestoreFile() {
         File return_;
-        String dataDir = config.getDataFileLocation();
+        String dataDir = config.getRestoreDataLocation();
         switch (type) {
             case SECONDARY_INDEX_V2:
-                String restoreFileName =
-                        PATH_JOINER.join(dataDir, keyspace, columnFamily, indexDir, fileName);
-                return_ = new File(restoreFileName);
+                return_ =
+                        new File(
+                                PATH_JOINER.join(
+                                        dataDir, keyspace, columnFamily, indexDir, fileName));
                 break;
             case META_V2:
                 return_ = new File(PATH_JOINER.join(config.getDataFileLocation(), fileName));
--- a/priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
+++ b/priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
@@ -21,6 +21,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.netflix.priam.backupv2.IMetaProxy;
 import com.netflix.priam.utils.DateUtil;
+import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -141,19 +142,18 @@ public class BackupRestoreUtil {
      * @param dataDirectory the location of the data folder.
      * @param monitoringFolder folder where cassandra backup's are configured.
      * @return Set of the path(s) containing the backup folder for each columnfamily.
-     * @throws Exception incase of IOException.
+     * @throws IOException
      */
     public static ImmutableSet<Path> getBackupDirectories(
-            String dataDirectory, String monitoringFolder) throws Exception {
+            String dataDirectory, String monitoringFolder) throws IOException {
         ImmutableSet.Builder<Path> backupPaths = ImmutableSet.builder();
         Path dataPath = Paths.get(dataDirectory);
         if (Files.exists(dataPath) && Files.isDirectory(dataPath))
             try (DirectoryStream<Path> directoryStream =
-                    Files.newDirectoryStream(dataPath, path -> Files.isDirectory(path))) {
+                    Files.newDirectoryStream(dataPath, Files::isDirectory)) {
                 for (Path keyspaceDirPath : directoryStream) {
                     try (DirectoryStream<Path> keyspaceStream =
-                            Files.newDirectoryStream(
-                                    keyspaceDirPath, path -> Files.isDirectory(path))) {
+                            Files.newDirectoryStream(keyspaceDirPath, Files::isDirectory)) {
                         for (Path columnfamilyDirPath : keyspaceStream) {
                             Path backupDirPath =
                                     Paths.get(columnfamilyDirPath.toString(), monitoringFolder);
--- a/priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
+++ b/priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
@@ -18,9 +18,13 @@
 package com.netflix.priam.backup;
 
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.netflix.priam.backupv2.IMetaProxy;
 import com.netflix.priam.utils.DateUtil;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -129,5 +133,37 @@ public class BackupRestoreUtil {
                     && (includeFilter.get(keyspace).isEmpty()
                             || includeFilter.get(keyspace).contains(columnFamilyName)));
         return false;
+    }
+
+    /**
+     * Get all the backup directories for Cassandra.
+     *
+     * @param dataDirectory the location of the data folder.
+     * @param monitoringFolder folder where cassandra backup's are configured.
+     * @return Set of the path(s) containing the backup folder for each columnfamily.
+     * @throws Exception incase of IOException.
+     */
+    public static ImmutableSet<Path> getBackupDirectories(
+            String dataDirectory, String monitoringFolder) throws Exception {
+        ImmutableSet.Builder<Path> backupPaths = ImmutableSet.builder();
+        Path dataPath = Paths.get(dataDirectory);
+        if (Files.exists(dataPath) && Files.isDirectory(dataPath))
+            try (DirectoryStream<Path> directoryStream =
+                    Files.newDirectoryStream(dataPath, path -> Files.isDirectory(path))) {
+                for (Path keyspaceDirPath : directoryStream) {
+                    try (DirectoryStream<Path> keyspaceStream =
+                            Files.newDirectoryStream(
+                                    keyspaceDirPath, path -> Files.isDirectory(path))) {
+                        for (Path columnfamilyDirPath : keyspaceStream) {
+                            Path backupDirPath =
+                                    Paths.get(columnfamilyDirPath.toString(), monitoringFolder);
+                            if (Files.exists(backupDirPath) && Files.isDirectory(backupDirPath)) {
+                                backupPaths.add(backupDirPath);
+                            }
+                        }
+                    }
+                }
+            }
+        return backupPaths.build();
     }
 }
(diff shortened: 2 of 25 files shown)


priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
```diff
@@ -147,15 +146,13 @@ protected String match(Date start, Date end) {
...
+        String dataDir = config.getRestoreDataLocation();
         switch (type) {
-            case CL:
-                return_ = new File(PATH_JOINER.join(config.getBackupCommitLogLocation(), fileName));
-                break;
             case SECONDARY_INDEX_V2:
-                String restoreFileName =
-                        PATH_JOINER.join(dataDir, keyspace, columnFamily, indexDir, fileName);
-                return_ = new File(restoreFileName);
+                return_ =
```
jrwest: While we are in here, consider renaming this variable, its not idiomatic java

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -144,4 +135,35 @@ public final boolean isFiltered(String keyspace, String columnFamilyDir) {
...
+     * @param dataDirectory the location of the data folder.
+     * @param monitoringFolder folder where cassandra backup's are configured.
+     * @return Set of the path(s) containing the backup folder for each columnfamily.
+     * @throws IOException
+     */
+    public static ImmutableSet<Path> getBackupDirectories(
+            String dataDirectory, String monitoringFolder) throws IOException {
+        ImmutableSet.Builder<Path> backupPaths = ImmutableSet.builder();
+        Path dataPath = Paths.get(dataDirectory);
+        if (Files.exists(dataPath) && Files.isDirectory(dataPath))
```
jrwest: The `File.exists` call is redundant here: https://docs.oracle.com/javase/6/docs/api/java/io/File.html#isDirectory()

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -144,4 +135,35 @@ public final boolean isFiltered(String keyspace, String columnFamilyDir) {
...
+            try (DirectoryStream<Path> directoryStream =
+                    Files.newDirectoryStream(dataPath, Files::isDirectory)) {
+                for (Path keyspaceDirPath : directoryStream) {
+                    try (DirectoryStream<Path> keyspaceStream =
+                            Files.newDirectoryStream(keyspaceDirPath, Files::isDirectory)) {
+                        for (Path columnfamilyDirPath : keyspaceStream) {
+                            Path backupDirPath =
+                                    Paths.get(columnfamilyDirPath.toString(), monitoringFolder);
```
jrwest: I believe you can use `columnFamilyDirPath.resolve(monitoringFolder)` here instead of converting back and forth between string and path. Reference: https://docs.oracle.com/javase/8/docs/api/index.html?java/nio/file/Path.html

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
```diff
@@ -144,4 +135,35 @@ public final boolean isFiltered(String keyspace, String columnFamilyDir) {
...
+                            if (Files.exists(backupDirPath) && Files.isDirectory(backupDirPath)) {
```
jrwest: Same comment here that the `File.exists` call is redundant.

priam/src/main/java/com/netflix/priam/backupv2/BackupVerificationTask.java
```diff
@@ -74,9 +74,7 @@ public void execute() throws Exception {
             return;
         }
 
-        if (instanceState.getRestoreStatus() != null
-                && instanceState.getRestoreStatus().getStatus() != null
-                && instanceState.getRestoreStatus().getStatus() == Status.STARTED) {
+        if (instanceState.isRestoring()) {
```
jrwest, line 77: I like this cleanup 🙌

priam/src/main/java/com/netflix/priam/backupv2/IMetaProxy.java
```diff
@@ -78,7 +78,7 @@ public interface IMetaProxy {
      * @param dateRange the time period to scan in the remote file system for incremental files.
      * @return iterator containing the list of path on the remote file system satisfying criteria.
```
jrwest, line 79: update javadoc if you are going to keep `ImmutableList`

priam/src/main/java/com/netflix/priam/backupv2/IMetaProxy.java
```diff
@@ -78,7 +78,7 @@ public interface IMetaProxy {
...
      */
-    Iterator<AbstractBackupPath> getIncrementals(DateUtil.DateRange dateRange);
+    ImmutableList<AbstractBackupPath> getIncrementals(DateUtil.DateRange dateRange);
```
jrwest, line 81: mostly curious: why this change? the iterator gives us flexibility that the data could be "streamed" where as this locks us in to determining all the incrementals up front. Likely not an issue but something I noticed.

jrwest: Looking at the implementation we have converted from something that would get called on each iteration to something that gets called all during the execution of this method. Not sure thats a problem or not but wanted to call it out.

priam/src/main/java/com/netflix/priam/backupv2/MetaV2Proxy.java
```diff
@@ -136,10 +129,7 @@ public List<AbstractBackupPath> findMetaFiles(DateUtil.DateRange dateRange) {
             AbstractBackupPath abstractBackupPath = abstractBackupPathProvider.get();
             abstractBackupPath.parseRemote(iterator.next());
             logger.debug("Meta file found: {}", abstractBackupPath);
-            if (abstractBackupPath.getLastModified().toEpochMilli()
-                            >= dateRange.getStartTime().toEpochMilli()
-                    && abstractBackupPath.getLastModified().toEpochMilli()
-                            <= dateRange.getEndTime().toEpochMilli()) {
+            if (dateRange.contains(abstractBackupPath.getLastModified())) {
```
jrwest, line 132: Really like this sort of cleanup as well.

priam/src/main/java/com/netflix/priam/config/IConfiguration.java
```diff
@@ -403,6 +398,10 @@ default String getRestoreSnapshot() {
         return StringUtils.EMPTY;
     }
 
+    default String getRestoreDataLocation() {
+        return getCassandraBaseDirectory() + "/restore";
```
jrwest, line 402: To confirm my understanding on our systems this would be something like `/mnt/data/cassandra/restore`? If so, +1

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
```diff
@@ -202,4 +211,56 @@ public List<Map<String, String>> gossipInfo() throws Exception {
...
+    }
+
+    private List<String> importData(String keyspace, String table, String source)
+            throws IOException {
+        try (JMXNodeTool nodeTool = JMXNodeTool.instance(configuration)) {
+            return nodeTool.importNewSSTables(
+                    keyspace,
+                    table,
+                    ImmutableSet.of(source),
+                    false /* resetLevel */,
```
jrwest, line 252: These defaults LGTM, do we want to add options to override them?

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
```diff
@@ -202,4 +211,56 @@ public List<Map<String, String>> gossipInfo() throws Exception {
...
+    public List<String> importAll(String srcDir) throws IOException {
+        List<String> failedImports = new ArrayList<>();
+        if (CassandraMonitor.hasCassadraStarted()) {
+            for (Path tableDir : BackupRestoreUtil.getBackupDirectories(srcDir, "")) {
+                String keyspace = tableDir.getParent().getFileName().toString();
+                String table = tableDir.getFileName().toString().split("-")[0];
+                failedImports.addAll(importData(keyspace, table, tableDir.toString()));
+            }
+        } else {
+            recursiveMove(Paths.get(srcDir), Paths.get(configuration.getDataFileLocation()));
```
jrwest: Since the move is non-atomic I can see a weird edge case where we start Cassandra while the move is happening (and we have like 10,000+ files). Should we have some sort of "fence" that tells the subsystem that starts Cassandra it needs to wait while this happens?

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
```diff
@@ -202,4 +211,56 @@ public List<Map<String, String>> gossipInfo() throws Exception {
...
+            if (!destination.toFile().mkdirs()) {
+                throw new IOException("Failed creating " + destination);
+            }
+        }
+        try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(source)) {
+            for (Path path : directoryStream) {
+                if (Files.isRegularFile(path)) {
+                    Files.move(path, destination.resolve(path.getFileName()));
+                } else if (Files.isDirectory(path)) {
+                    recursiveMove(path, destination.resolve(path.getFileName()));
```
jrwest: If am reading this right, if the source path we are currently iterating over is a directory we recurse by moving the files in that directory into `[destination/directory name]`. Cant we just move the entire directory instead of recursing then?

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
```diff
@@ -202,4 +211,56 @@ public List<Map<String, String>> gossipInfo() throws Exception {
...
+                } else {
+                    throw new IOException("Failed determining type of inode is " + path);
```
jrwest: Do we need this error or just skip/log? Likely an edge case but do we really want to fail if thats the case?

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
```diff
@@ -95,161 +89,94 @@ public AbstractRestore(
...
                                 new Date(dateRange.getStartTime().toEpochMilli()));
                 instanceIdentity.getInstance().setToken(restoreToken.toString());
             }
-
-            // Stop cassandra if its running
-            stopCassProcess();
-
-            // Cleanup local data
-            File dataDir = new File(config.getDataFileLocation());
-            if (dataDir.exists() && dataDir.isDirectory()) FileUtils.cleanDirectory(dataDir);
```
jrwest, line 199: Question about this: I know we don't need to stop `cass` anymore because we use `import` but I was surprised to see we no longer clean the directory. Is this for the case where we want to restore a subset of the data but keep whats there? Do we still need the option to clean the directory for some restores?

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
```diff
@@ -95,161 +89,94 @@ public AbstractRestore(
...
-            instanceState.setRestoreStatus(Status.FINISHED);
-
-            // Start cassandra if restore is successful.
-            if (!config.doesCassandraStartManually()) cassProcess.start(true);
-            else
-                logger.info(
-                        "config.doesCassandraStartManually() is set to True, hence Cassandra needs to be started manually ...");
+            List<String> failedImports = cassOps.importAll(config.getRestoreDataLocation());
+            if (!failedImports.isEmpty()) {
+                instanceState.endRestore(Status.FAILED, LocalDateTime.now());
```
jrwest, line 170: do we want to use the list of failed imports somewhere to log what failed and be able to resume? instead of starting over from the beginning? 

Also if we only need the boolean why return the list? I would rather us use the list but if we are not going to I mean.

priam/src/main/java/com/netflix/priam/backup/AbstractBackupPath.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/backup/BackupRestoreUtil.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
mattl-netflix: In the new model using walkFileTree() we always continue.

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
mattl-netflix: The common cases for restores are:
1. Restores into empty clusters (e.g., red-black)
2. Restored into live clusters (e.g., emergencies)

In neither case do we want any data to be deleted. I would prefer that in outlier cases where that is required that there is a separate explicit job or command that the operator needs to run to best manage expectations. It is scary today that we could configure Priam such that a restart deletes all of the data.

priam/src/main/java/com/netflix/priam/restore/AbstractRestore.java
mattl-netflix: This is a good point. In case we don't specify --copy-data and Cassandra's behavior is to remove the origin files regardless of successful import, we want a log of what failed so we can at least get a sense of the scope of what is missing. I will log the failures.

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
mattl-netflix: My goal here was for the the operation to be strictly additive. That is, if there are already files present I want to ensure that for them to get overwritten that the operator does something explicit.

Note that to that end I have changed the policy regarding whether to replace files if they are already present in a follow-up commit.

priam/src/main/java/com/netflix/priam/connection/CassandraOperations.java
mattl-netflix: Yes we should. I will need to address this in a follow-up, however. This is analogous to Cassandra getting started accidentally before the full download is complete the in the model we use today. Which hasn't happened yet, knock on wood. It might requires some rethinking about consolidating Cassandra's remediation policy in our environment. That is, the between Priam and systemd.

PR #1098, opened 2024-06-05
Prefer replacing dead nodes with the same IP.

mattl-netflix:


priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java
```diff
@@ -136,7 +136,13 @@ public PriamInstance retriableCall() throws Exception {
                                 .filter(i -> !racInstanceIds.contains(i.getInstanceId()))
                                 .collect(Collectors.toList());
                 Optional<PriamInstance> candidate =
-                        instances.stream().filter(i -> !isNew(i)).findFirst();
+                        instances
```
clohfink, line 139: NP: Can ignore this but could do the isNew and same ip in same stream with a reduce like 

```Optional<PriamInstance> candidate = instances.stream()
    .reduce((current, next) -> {
        if (myInstanceInfo.getPrivateIP().equals(next.getHostIP())) {
            return next;
        } else if (!isNew(next) && !current.isPresent()) {
            return next;
        }
        return current;
    });```

**Review by clohfink — APPROVED** [review]

**Review by chengw-netflix — APPROVED** [review]

File: priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java (hunk 1 above)
Comment (mattl-netflix) [reply priam/src/main/java/com/netflix/priam/identity/token/TokenRetriever.java:139]: Ahh the oft-neglected reduce! Though recall we are dealing with PriamInstances not Optionals so the !current.isPresent() check doesn't work as expected.  Instead the logic should be:

```java
filter(i -> !isNew(i))
.reduce(
    (current, next) ->
        myInstanceInfo.getPrivateIP().equals(next.getHostIP()) 
            ? next
            : current);
```

PR #1103, opened 2024-09-27
Do not delete backups if only the compression type differs.

mattl-netflix:


priam/src/main/java/com/netflix/priam/backupv2/BackupTTLTask.java
```diff
@@ -260,18 +261,18 @@ public static TaskTimer getTimer(
...
-                                                                    null)));
+            for (ColumnFamilyResult.SSTableResult sstable : columnfamilyResult.getSstables()) {
+                for (FileUploadResult component : sstable.getSstableComponents()) {
+                    filesInMeta.put(removeCompressionPart(component.getBackupPath()), null);
+                }
+            }
+        }
+    }
+
+    private String removeCompressionPart(String backupPath) {
```
clohfink: this can have some easy unit tests, would be helpful documentation a bit too on some expected paths look like

clohfink (approved):

PR #1117, opened 2025-08-09
Reusable buffer for backup uploading

clohfink:
Metric | byte[] impl | ByteBuffer impl | Change
-- | -- | -- | --
Throughput (ops/ms) | 0.414 (+/- 0.083) | 0.890 (+/- 0.058) | 2.15x faster
Throughput (ops/s) | 414 | 890 | +476 ops/s
Alloc per op (B/op) | 3,092,129.6 B/op | 2,421.4 B/op | 1277x less
Alloc per op (MiB/op) | ~2.95 MiB/op | ~0.0023 MiB/op | ~99.92% reduction
GC alloc rate (MB/s) | 1,043.3 | 1.76 | 593x less
GC count | 12 | 1 | 91.7% fewer GCs
GC time (ms) | 14 | 5 | 64.3% less time in GC


mattl-netflix:
This was committed in a later PR.

PR #1118, opened 2025-08-22
Warmup objectCache with the latest validated snapshot files to reduce HEAD requests to AWS

chengw-netflix:
On restart the upload cache is not warmed. Across the fleet this leads to a six-figure expense on each rollout due to the extra S3 HEAD calls that need to be made when uploading. Rather, we should call the backup metadata API to determine the most recent valid meta file and warm the cache using the object keys it contains within.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/aws/S3FileSystem.java
+++ b/priam/src/main/java/com/netflix/priam/aws/S3FileSystem.java
@@ -21,10 +21,9 @@ import com.amazonaws.services.s3.S3ResponseMetadata;
 import com.amazonaws.services.s3.model.*;
 import com.google.common.base.Preconditions;
 import com.netflix.priam.aws.auth.IS3Credential;
-import com.netflix.priam.backup.AbstractBackupPath;
-import com.netflix.priam.backup.BackupRestoreException;
-import com.netflix.priam.backup.DynamicRateLimiter;
-import com.netflix.priam.backup.RangeReadInputStream;
+import com.netflix.priam.backup.*;
+import com.netflix.priam.backupv2.IMetaProxy;
+import com.netflix.priam.backupv2.MetaV2Proxy;
 import com.netflix.priam.compress.ChunkedStream;
 import com.netflix.priam.compress.CompressionType;
 import com.netflix.priam.compress.ICompression;
@@ -67,8 +66,10 @@ public class S3FileSystem extends S3FileSystemBase {
             BackupMetrics backupMetrics,
             BackupNotificationMgr backupNotificationMgr,
             InstanceInfo instanceInfo,
-            DynamicRateLimiter dynamicRateLimiter) {
-        super(pathProvider, compress, config, backupMetrics, backupNotificationMgr);
+            DynamicRateLimiter dynamicRateLimiter,
+            IBackupStatusMgr backupStatusMgr,
+            IMetaProxy metaV2Proxy) {
+        super(pathProvider, compress, config, backupMetrics, backupNotificationMgr, backupStatusMgr, metaV2Proxy);
         s3Client =
                 AmazonS3Client.builder()
                         .withCredentials(cred.getAwsCredentialProvider())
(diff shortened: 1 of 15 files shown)


priam/src/main/java/com/netflix/priam/resources/BackupServletV2.java
```diff
@@ -120,6 +119,13 @@ public Response clearCache() throws Exception {
         return Response.ok(REST_SUCCESS, MediaType.APPLICATION_JSON).build();
     }
 
+    @GET
```
mattl-netflix: Please remove this.

Why? We're adding a method to a public API without the attendant telemetry that tells us who will get broken if we remove it. Since there are no plans to actually call this today, we should refrain from putting it in until there are.

priam/src/test/java/com/netflix/priam/resources/BackupServletV2Test.java
mattl-netflix, line 1: You can revert this file if you remove the REST API command.

priam/src/main/java/com/netflix/priam/backup/IBackupFileSystem.java
```diff
@@ -175,4 +175,11 @@ default boolean checkObjectExists(Path remotePath) {
 
     /** Clear the cache for the backup file system, if any. */
     void clearCache();
+
+    /**
+     * Warm up cache with the latest validated backup if any.
+     *
+     * @return the number of files added to cache.
+     * */
+    int warmupCache() throws Exception;
```
mattl-netflix: Instead of putting all of the logic into the file system abstraction prefer instead to make the addObjectCache(Path path) method public and part of the IBackupFileSystem API instead of `warmupCache()` as is done here.

We will instead put the meat of the logic either in SnapshotMetaTask directly or in BackupHelper.

This will avoid the circular dependency in which the metaProxy depends on the file system and vice versa.

priam/src/main/java/com/netflix/priam/aws/S3FileSystem.java
```diff
@@ -67,8 +66,10 @@ public S3FileSystem(
             BackupMetrics backupMetrics,
             BackupNotificationMgr backupNotificationMgr,
             InstanceInfo instanceInfo,
-            DynamicRateLimiter dynamicRateLimiter) {
-        super(pathProvider, compress, config, backupMetrics, backupNotificationMgr);
+            DynamicRateLimiter dynamicRateLimiter,
+            IBackupStatusMgr backupStatusMgr,
```
mattl-netflix: Adding these dependencies to all of these IBackupFileSystem implementations is a code smell. It indicates to me that we are putting this low in the stack.

Instead, revert the change to the AbstractFileSystem constructor and the analogous ones in its subclasses.

priam/src/main/java/com/netflix/priam/backup/AbstractFileSystem.java
```diff
@@ -321,4 +329,42 @@ public int getDownloadTasksQueued() {
     public void clearCache() {
         objectCache.invalidateAll();
     }
+
+    @Override
+    public int warmupCache() throws Exception {
```
mattl-netflix: If you put this in BackupHelperImpl the logic resolves to:

```Java
    @Override
    public void warmupCache(IMetaProxy metaProxy, IBackupStatusMgr snapshotStatusMgr) throws Exception {
        Instant now = Instant.now();
        DateUtil.DateRange dateRange = new DateUtil.DateRange(now.minus(1, ChronoUnit.DAYS), now);
        Optional<String> snapshotLocation = snapshotStatusMgr.getLatestBackupMetadata(dateRange)
                .stream()
                .filter(backupMetadata -> backupMetadata.getLastValidated() != null)
                .max(Comparator.comparing(BackupMetadata::getStart))
                .map(BackupMetadata::getSnapshotLocation);
        if (snapshotLocation.isPresent()) {
            AbstractBackupPath backupPath = pathFactory.get();
            backupPath.parseRemote(snapshotLocation.get());
            BackupRestoreUtil.getMostRecentSnapshotPaths(backupPath, metaProxy, pathFactory)
                    .stream()
                    .map(abp -> Paths.get(abp.getRemotePath()))
                    .forEach(fs::addToCache);
        }
    }
```

mattl-netflix (changes requested):
We'll need to call this on startup. To do that, call `snapshotMetaTask.warmCache()` in BackupV2Service. Be sure it is done before any call to uploadFiles or to schedule the task.

That method can be implemented in SnapshotMetaTask like so:

```Java
    public void warmCache() throws Exception {
        backupHelper.warmupCache(metaProxy, snapshotStatusMgr);
    }
```

mattl-netflix (approved):

PR #1121, opened 2025-10-01
Deprecating SimpleDB

sinchanah-dot:


mattl-netflix (approved):
This effectively marks the end of Netflix's support of Priam as a standalone, self-contained application. We'll need to remember to at least state that in the release notes.

PR #1122, opened 2025-10-14
Persist validation markers to S3 for UCC.

mattl-netflix:
This persists an alternative location for backup metadata in S3 in case of unavailability of the primary source of truth. Tests are in progress and will be added in a follow-up commit. I have confirmed it WAI in practice.

### DIFF
--- a/priam/src/main/java/com/netflix/priam/backupv2/SnapshotValidationMarkerWriter.java
+++ b/priam/src/main/java/com/netflix/priam/backupv2/SnapshotValidationMarkerWriter.java
@@ -1,0 +1,53 @@
+package com.netflix.priam.backupv2;
+
+import com.netflix.priam.aws.RemoteBackupPath;
+import com.netflix.priam.backup.AbstractBackupPath;
+import com.netflix.priam.backup.IBackupFileSystem;
+import com.netflix.priam.config.IConfiguration;
+import com.netflix.priam.identity.config.InstanceInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+public class SnapshotValidationMarkerWriter {
+    private static final Logger logger = LoggerFactory.getLogger(SnapshotValidationMarkerWriter.class);
+    private static final String METADATA_PREFIX = "metadata";
+    private final IConfiguration config;
+    private final InstanceInfo instanceInfo;
+    private final Provider<AbstractBackupPath> pathProvider;
+    private final IBackupFileSystem fs;
+
+    @Inject
+    public SnapshotValidationMarkerWriter(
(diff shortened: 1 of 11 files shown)


priam/src/main/java/com/netflix/priam/backupv2/SnapshotValidationMarkerWriter.java
```diff
@@ -0,0 +1,53 @@
...
+import com.netflix.priam.backup.IBackupFileSystem;
+import com.netflix.priam.config.IConfiguration;
+import com.netflix.priam.identity.config.InstanceInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+public class SnapshotValidationMarkerWriter {
```
clohfink: NP: use of Verification instead of Validation other places, rename to maintain consistent terminology

priam/src/main/java/com/netflix/priam/backupv2/SnapshotVerificationMarkerWriter.java
```diff
@@ -0,0 +1,53 @@
...
+        String key = AbstractBackupPath.PATH_JOINER.join(
+                METADATA_PREFIX,
+                RemoteBackupPath.prependHash(path.getClusterName()),
+                path.getLastModified().toEpochMilli(),
+                instanceInfo.getRac(),
+                path.getToken());
+        try {
+            fs.putObject(config.getBackupPrefix(), key, remotePath);
+        } catch (Exception e) {
+            logger.error("Failed to put snapshot verification marker. bucket: {}, key: {}, value: {}, message: {}",
```
clohfink, line 53: This could use a retry since it doesnt provide any metrics or anything to notify and we expect S3 puts to fail sometimes.

clohfink: also a success/failure metric might be good to add

clohfink (approved):
+1 with some changes you can ignore if disagree

priam/src/main/java/com/netflix/priam/backupv2/SnapshotValidationMarkerWriter.java
mattl-netflix: Done.

priam/src/main/java/com/netflix/priam/backupv2/SnapshotVerificationMarkerWriter.java
mattl-netflix: Added retries. We can metrics implicitly when we centralize validation and include these in the checks.

IMO we can justify stacking a bet here because this is meant as a fallback rather than the main mechanism to get these data.

PR #1125, opened 2025-10-24
Bug fix for object cache warmup. 

chengw-netflix:


mattl-netflix (approved):
Approval assumes this has been tested in practice.

PR #1127, opened 2025-12-04
Set the marker to the last key when delimiter is null.

mattl-netflix:
This prevents endless paginating through list results.


sinchanah-dot:
The original code was calling objectListing.nextMarker() unconditionally for pagination, but S3 only populates nextMarker when a delimiter is specified. Without a delimiter, nextMarker() would be null or empty, causing the iterator to either fail to paginate properly or get stuck in an endless loop paginating through the same results.

This change looks good to me!

